PR #8, opened 2015-12-05
Added functionality to handle Result in node compiler

lukewestby:
I really like the approach you're taking here and wanted to help out :smile:
Even if you'd rather not accept the changes I would love any feedback if you have time. Thanks!

- Added a record type to represent the `Result` from `prettyPrint` in node
- Added a helper function to make that conversion
- Updated node compiler to exit with code 1 if the `Result` is `Err`
- Updated example elm file to use `Css` correctly

PR #26, opened 2015-12-22
Moving compiler output changes to master

lukewestby:
From #8 

Thanks @rtfeldman!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 elm-stuff
 node_modules
+output.css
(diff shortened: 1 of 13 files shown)


rtfeldman:
Thanks so much for this! :heart_eyes_cat:

lukewestby:
CSS-in-Elm is a critical feature for the ecosystem IMO so I'm happy to help out :+1:

lukewestby:
I'm keeping a mental list of "things that are already available in JS but not in Elm and that's why we aren't going to switch" and trying to bring them to Elm to eliminate all excuses. This package will be wonderful to have when it's ready

PR #57, opened 2016-01-03
Add a bunch of elements

lukewestby:
I tried to go with the strategy of "every element I've used or seen used before". There are a lot of elements I've never even heard of :flushed:

Let me know if you think this covers enough ground, I'll be happy to add more!


lukewestby:
Oh shoot, looks like there's ambiguity between the `em` units function and the `<em>` element. Thoughts?

joshmh:
Yeah, tough one. A similar problem is addressed here by adding an apostrophe, e.g., **em'**: https://github.com/adam-r-kowalski/elm-style.

Another approach, which I prefer, is to just rename the subjectively less important one. In this case, I'd say the unit em is more important, and I'd rename the tag. We'd have to figure out a good naming convention in this case: emTag is a possibility.

lukewestby:
I wonder, if the intention of moving the elements into a separate module was to prevent clashes with elm-html, would it make sense to just leave `em` as is and update the test fixtures file to look something like:

```elm
module FixturesCss (..) where

import Css exposing (..)
import Css.Elements as El

unstyledDiv : Stylesheet String a b c
unstyledDiv =
    stylesheet { name = "" }
        $ El.div


divWidthHeight : Stylesheet String a b c
divWidthHeight =
    stylesheet { name = "" }
        $ El.div
            ~ width (pct 32)
            ~ height (px 50)

-- ...
```

I'm on board with `emTag` as well. The only thing I'm not sure about is `em'`, since I was under the impression that `'` was best limited to names that clash with keywords, like `type'` and `main'` from elm-html, but I could of course be totally wrong. @rtfeldman what do you think?

rtfeldman:
I'd say just leave `em` out of Elements. `em` the length unit gets used all the time in stylesheets, but how often do you want to restyle an `em` tag? Seems worth leaving it out of the library in order to avoid the collision; you can always use `$= "em"` if you really need to style one.

lukewestby:
Anything y'all think is missing or anything in there that seems unnecessary?

rtfeldman:
Nope, love it!

rtfeldman:
Thanks a ton @lukewestby! :tada:

lukewestby:
Woo!!! I'm excited for how well this is all coming along :smile:

PR #58, opened 2016-01-05
Flex box

NicholasGW:
Hey Richard,

Took a stab at this to try and get better with Elm, but ran into a hurdle. This seems to work for the "length" values quoted in the spec except for content. I tried to mirror something similar to `lengthOrAutoOrMinMaxDimension` but couldn't really figure out how those work and what "compatible" is doing. If you could provide some guidance I'd be pumped to try and add some more!

Essentially, if I'm understanding right, I'd need to make a new function 'content' so that `flexBasis content`  would output `flex-basis: content` ? But that would require a new type alias like `LengthOrAutoOrContent`?  
  
Thanks, this repo is great for CSS and learning a bit of elm.

Nick


NicholasGW:
Ah I see now it's the extensible record format, wasn't familiar with that. I'll keep digging at it :)

NicholasGW:
New commit might make it make sense. I added a flexBasis alias, added it to explicit length, auto, initial, inherit, unset, then added a content function to make the value content. Is this more in line with how things should be added :)?

rtfeldman:
@NicholasGW regarding [your `flexFlow` question](https://www.reddit.com/r/elm/comments/3zrpz8/help_need_function_with_2_possible_inputs_in/) - the convention I've been following would be to have `flexFlow` and `flexFlow2`, which would work similarly to [`margin` and `margin2`](https://github.com/rtfeldman/elm-css/blob/master/src/Css.elm#L1892-L1913).

Does that make sense?

NicholasGW:
@rtfeldmam Thanks for the comments! That's what I was thinking for flex flow, but I couldn't get around that flexflow1 could take either FlexDirection or FlexWrap compatible... With margin and text-shadow it makes perfect sense to me because they're either all the same type and CSS knows which is which or they have a well defined order... Unless I'm missing something :p

rtfeldman:
Oh, got it...this should do the trick:

1. Introduce `flexDirectionOrWrap : Compatible`
2. Have everything with a `flexDirection : Compatible` field also have a `flexDirectionOrWrap : Compatible` field
3. Have everything with a `flexWrap : Compatible` field also have a `flexDirectionOrWrap : Compatible` field
4. Have `flexFlow2` take `{ a | flexDirectionOrWrap : Compatible } -> { b | flexDirectionOrWrap : Compatible }` as arguments.

NicholasGW:
Ah awesome, makes sense! Should be able to finish flex-box tonight then.

Cheers!

NicholasGW:
Also added order, but realized justify-content and align-content aren't there either :).

rtfeldman:
Looks great @NicholasGW, thanks a ton! :heart_eyes_cat:

NicholasGW:
My pleasure!

Issue #20, opened 2015-12-21
Font keywords


rtfeldman:
These are keywords necessary to cover Font

- [ ] [`@character-variant`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-feature-values#@character-variant)
- [ ] [`character-variant()`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-alternates#character-variant())
- [ ] [`font`](https://developer.mozilla.org/en-US/docs/Web/CSS/font)
- [ ] [`@font-face`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face)
- [ ] [`font-family`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family)
- [ ] [`font-feature-settings`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-feature-settings)
- [ ] [`@font-feature-values`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-feature-values)
- [ ] [`font-kerning`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-kerning)
- [ ] [`font-language-override`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-language-override)
- [ ] [`font-size`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size)
- [ ] [`font-size-adjust`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size-adjust)
- [ ] [`font-stretch`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-stretch)
- [ ] [`font-style`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-style)
- [ ] [`font-synthesis`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-synthesis)
- [ ] [`font-variant`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant)
- [ ] [`font-variant-alternates`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-alternates)
- [ ] [`font-variant-caps`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-caps)
- [ ] [`font-variant-east-asian`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-east-asian)
- [ ] [`font-variant-ligatures`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-ligatures)
[...]

Issue #14, opened 2015-12-21
Miscellaneous keywords


rtfeldman:
These are keywords that are in the spec but not mentioned in other issues. Some of them should probably be moved to https://github.com/rtfeldman/elm-css/issues/12

- [ ] [`align-content`](https://developer.mozilla.org/en-US/docs/Web/CSS/align-content)
- [x] [`all`](https://developer.mozilla.org/en-US/docs/Web/CSS/all)
- [ ] [`attr`](https://developer.mozilla.org/en-US/docs/Web/CSS/attr)
- [ ] [`backface-visibility`](https://developer.mozilla.org/en-US/docs/Web/CSS/backface-visibility)
- [ ] [`block-size`](https://developer.mozilla.org/en-US/docs/Web/CSS/block-size)
- [ ] [`box-decoration-break`](https://developer.mozilla.org/en-US/docs/Web/CSS/box-decoration-break)
- [ ] [`box-shadow`](https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow)
- [ ] [`box-sizing`](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing)
- [ ] [`break-after`](https://developer.mozilla.org/en-US/docs/Web/CSS/break-after)
- [ ] [`break-before`](https://developer.mozilla.org/en-US/docs/Web/CSS/break-before)
- [ ] [`break-inside`](https://developer.mozilla.org/en-US/docs/Web/CSS/break-inside)
- [ ] [`calc()`](https://developer.mozilla.org/en-US/docs/Web/CSS/calc())
- [ ] [`caption-side`](https://developer.mozilla.org/en-US/docs/Web/CSS/caption-side)
- [ ] [`@charset`](https://developer.mozilla.org/en-US/docs/Web/CSS/@charset)
- [ ] [`circle()`](https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape#circle)
- [ ] [`clear`](https://developer.mozilla.org/en-US/docs/Web/CSS/clear)
- [ ] [`clip`](https://developer.mozilla.org/en-US/docs/Web/CSS/clip)
- [ ] [`clip-path`](https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path)
- [ ] [`color`](https://developer.mozilla.org/en-US/docs/Web/CSS/color)
- [ ] [`content`](https://developer.mozilla.org/en-US/docs/Web/CSS/content)
[...]

PR #64, opened 2016-01-19
Issue #14 and #20

wmjoubert:
Added line-height, font-family, font-style, font-size, font-weight and font-variant together with various supporting keyword such as bold, serif, etc. They're mostly complete, there might be some legal CSS not accounted for. 

Commented the unsupported medium function and add medium as a valid font weight attribute.


src/Css.elm
```diff
@@ -4128,9 +4647,9 @@ fall back on something like this:
         $ button
             ~ ( "border-left", "thin" )
 -}
-medium : IntentionallyUnsupportedPleaseSeeDocs
-medium =
-  IntentionallyUnsupportedPleaseSeeDocs
+-- medium : IntentionallyUnsupportedPleaseSeeDocs
+-- medium =
+--     IntentionallyUnsupportedPleaseSeeDocs
```
rtfeldman: This is fine, although I'd say just delete it.

src/Css.elm
```diff
@@ -1037,6 +1126,22 @@ n num =
...
+      { value = numberToString num, fontWeight = Compatible }
+    else
+      if num < 100 then
+        { value = "100", fontWeight = Compatible }
+      else if num > 900 then
+        { value = "900", fontWeight = Compatible }
+      else
+        { value = numberToString <| round (num / 100) * 100
+        , fontWeight = Compatible }
+
```
rtfeldman: I think it would be more consistent to implement this using the existing `n` function (for bare numeric units) and a warning via `fontWeight` for invalid values.

For example:

`fontWeight (n 100)` yields:
```elm
{ key = "font-weight"
, value = "100"
, warnings = []
}
```

`fontWeight (n 22)` yields:
```elm
{ key = "font-weight",
, value = "22"
, warnings = [ "22 is not a valid font-weight value. Please see https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#Values" ]
}
```

src/Css.elm
```diff
@@ -3402,6 +3763,164 @@ media value =
...
+    prop1 "font-variant-ligatures"
+
+
+fontVariantCaps : FontVariantCaps a -> Mixin
+fontVariantCaps =
+    prop1 "font-variant-caps"
+
+{- TODO Handle
+    font-variant-numeric: oldstyle-nums tabular-nums stacked-fractions ordinal slashedZero
+-}
```
rtfeldman: Thanks for the `TODO`!

rtfeldman:
This is fantastic, thank you so much! :heart_eyes_cat: 

I added one comment about `w` but otherwise looks great!

wmjoubert:
The tests fail due to the change in at-rule semantics - how do you want to handle standalone at-rules?

rtfeldman:
Awesome, thank you so much for this!

PR #66, opened 2016-01-20
Adds hsl and hsla color values

lukewestby:
The `s` and `l` properties accept a percentage value in CSS, but these percentages are different than lengths and percentages higher than 100% are meaningless so I thought it made sense to just have them be numbers between 0 and 1.

Thanks!


rtfeldman:
Love it! :smile_cat: Thanks!

PR #77, opened 2016-01-30
Address TODO under importToString

wmjoubert:
Implemented it like this since I need it for the project I'm working on. Handles most common cases. Will add a test if you approve. I suspect `NameOrUri` should be moved to Css.elm and possibly renamed to `NameOrUrl` with an `Url` tag instead of `Uri` to be more inline with the CSS semantics.

### DIFF
--- a/src/Css/Preprocess.elm
+++ b/src/Css/Preprocess.elm
@@ -4,7 +4,7 @@ module Css.Preprocess (..) where
 the data structures found in this module.
 -}
 
-import Css.Structure as Structure exposing (mapLast, concatMapLast)
+import Css.Structure as Structure exposing (mapLast, concatMapLast, NameOrUri)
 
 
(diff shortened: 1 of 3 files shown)


rtfeldman:
Love it! Some thoughts:

* :+1: for changing it to `NameOrUrl`
* Would love some tests!
* `Structure` shouldn't depend on anything from `Css`, so defining it there makes sense—but I think having `Css` import and then re-export both `NameOrUrl` and its two constructors is a good idea.

Thanks for this! :smile:

wmjoubert:
Could not find a way to re-export a union so created `name` and `url` functions in `CSS` instead. If you prefer the tagged type and know how to re-export them please let me know. I'm also not entirely sure how you intended setting things like `imports` and `charset` on a `Stylesheet` - returning a modified record like in the added test at the moment.

Issue #16, opened 2015-12-21
Background keywords


rtfeldman:
These are keywords necessary to cover Background

- [ ] [`background`](https://developer.mozilla.org/en-US/docs/Web/CSS/background)
- [ ] [`background-attachment`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-attachment)
- [ ] [`background-blend-mode`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-blend-mode)
- [ ] [`background-clip`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip)
- [ ] [`background-color`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-color)
- [ ] [`background-image`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-image)
- [ ] [`background-origin`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-origin)
- [ ] [`background-position`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-position)
- [ ] [`background-repeat`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-repeat)
- [ ] [`background-size`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-size)

jwoudenberg:
I've got this one!

rtfeldman:
Awesome, thank you! :smile_cat:

rtfeldman:
I've decided to close all the `missing keywords` issues in favor of letting people make PRs based on what specific things they wish existed.

PR #80, opened 2016-02-07
Background keywords (#16) (WIP)

jwoudenberg:
**NOTE: I believe this is ready to merge! Please see my bottom-most comment.**

This PR is not yet ready to merge. I'm looking for some feedback to finish it and though this might be useful for discussion.

I've been working on the background keywords (issue #16) and have completed most of the grunt work. However, some issues remain that I'd appreciate some help with. I'm still pretty new to Elm, so please have some patience with me :).

### 1. `background-repeat`
The `background-repeat` property allows the value `round` which shadows `Basics.round`. It's unfortunate, but personally I think its best to leave it this way and have users type `Css.round` if they want to use it. The alternative of giving it another name I personally think would create more confusion then it solves. Do you agree?

### 2. `background-blend-mode`
The `background-blend-mode` property allows keywords `color` and `screen`. `color` I handled using the overloaded property trick. `screen` hoever is not a property but a media query, so I don't think that will work. I'm not sure how to proceed here. Could we someone extend the trick to accept media queries as well?

### 3. `background-position`
The `background-position` keyword allows a positional keyword like `left`, but also a length like `10 px`. I can use the overload trick to have it accept the keywords, but then it won't accept something like `(px 10)` anymore. I could use a hint to carry on :).

### 4. `background`
The `background` css property accepts any number of background related values in any order. What I could do is create a `background` key that takes all properties in a particular order, but I wonder how useful that would be compared to using the properties separately. Personally, I like the idea of being able to pass a record like so:

```elm
background =
  { repeat = repeatX
  , image = url 'http://example.com/elm.png
  }
```

Where you basically pass shorthand keys for the different background properties. Would that be a fitting solution?

### 5. multiple backgrounds
The background css properties allow multiple backgrounds to be defined. This is done by passing multiple values per property separated by comma's:

```css
background-repeat: repeatX, space repeat
```

Porting this notation style to Elm as literally as possible would mean all the background properties take lists. Personally, I do not like this idea very much for two reasons. First, it over-complicates the base case (using a single background). Secondly, a user would be forced to use the same version of each property for each background, so the above example would not work, because it would require us to be able to mix `backgroundRepeat` and `backgroundRepeat2` values in a single list.

I think that in this case it would be nicer to take whatever solution we settle on for point 4 above, and create a new property like `multipleBackgrounds` that takes a list of them. Arguably, grouping by values by background rather than key is an improvement over css, though I realize that might not be the most relevant argument here :).

Thanks for reading. Any feedback is greatly appreciated!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ install:
   - nvm use $TARGET_NODE_VERSION
   - node --version
   - npm --version
-  - cd test
+  - cd tests
   - npm install -g elm@$ELM_VERSION elm-test
   - npm install
   - git clone https://github.com/NoRedInk/elm-ops-tooling
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -529,12 +529,48 @@ module Css
         , listStyle
         , listStyle2
         , listStyle3
+        , backgroundRepeat
+        , backgroundRepeat2
+        , repeatX
+        , repeatY
+        , repeat
+        , space
+        , round
+        , noRepeat
+        , backgroundAttachment
+        , local
+        , backgroundBlendMode
+        , multiply
+        , overlay
+        , darken
+        , lighten
+        , colorDodge
+        , colorBurn
+        , hardLight
+        , softLight
+        , difference
+        , exclusion
+        , hue
+        , saturation
+        , luminosity
+        , screenBlendMode
+        , backgroundClip
+        , paddingBox
+        , backgroundImage
+        , url
+        , backgroundPosition
+        , backgroundPosition2
+        , backgroundOrigin
+        , backgroundSize
+        , backgroundSize2
+        , cover
+        , contain
         )
 
 {-| Functions for building stylesheets.
 
 # Misc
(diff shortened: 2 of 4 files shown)


jwoudenberg:
Hey there. I just wanted to remark that I've updated this PR to be up to date with master again and that any and all feedback would be really appreciated :smile:.

I'm not waiting for a detailed rebuttal to each and every point above or anything, anything small to get me going again would be appreciated. Or if the feeling is I'm in over my head and this PR is not going anywhere, that's ok to too, I'd just like to know :smile_cat:.

Thanks!

jwoudenberg:
Hello! I gave this some more love and I think it's ready to merge.

Details:
This is how I dealt with the points raised in my origin PR:
1. Not really a problem.
2. I gave the screen blend mode the not-so-nice name `screenBlendMode`. I personally do not care enough about this blendMode to invest much more time in it, but I'm open to any good ideas people might have :smile:.
3. I let the `backgroundPosition` take keywords and `backgroundPosition2` take length values. I think this fits how these keywords tends to be used. Let me know if you disagree!
4. Considering this property is specified as a list of values of different types, all optional and in any order, I don't think not supporting it in a strictly typed language is a big deal. Use the specific background properties instead.
5. This is not yet supported.

bgourlie:
I would like to see this land, currently in need of `backgroundPosition` support!

rtfeldman:
@jwoudenberg okay, finally got around to reviewing this - looks great! Sorry I took so long. 😅 

If you wouldn't mind one last rebase, I will gladly merge this!

jwoudenberg:
@rtfeldman no worries! Rebase is done.

rtfeldman:
Awesome! Thank you @jwoudenberg! 🎉

flipace:
This PR is great! Any idea when this (/the next version) will land?

rtfeldman:
I want to wait for https://github.com/rtfeldman/elm-css/pull/177 since it's a breaking change too. 🙂

Issue #13, opened 2015-12-21
Flexbox keywords


rtfeldman:
These are keywords necessary to cover Flexbox

- [ ] [`alignItems`](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items)
- [ ] [`alignSelf`](https://developer.mozilla.org/en-US/docs/Web/CSS/align-self)
- [ ] [`flex`](https://developer.mozilla.org/en-US/docs/Web/CSS/flex)
- [ ] [`flexBasis`](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis)
- [ ] [`flexDirection`](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction)
- [ ] [`flexFlow`](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-flow)
- [ ] [`flexGrow`](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow)
- [ ] [`flexShrink`](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink)
- [ ] [`flexWrap`](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-wrap)

NicholasGW:
Working on these currently, so nobody else picks them up :).

joshmh:
See https://github.com/rtfeldman/elm-css/pull/58

NicholasGW:
Thanks, that's my PR when I just asking Richard some questions :P Just though I'd post here now that I've done a few more in case nobody checked the PRs

joshmh:
Haha, right. Somehow I missed that it was you. Looking forward to this.

NicholasGW:
Having an issue with flex-flow, and likely flex.
  
With flex-flow, it can accept either flex-direction, flex-row, or both, in any order since the keywords are unique. 

My problem is, with most multi-parameter CSS properties, they have a defined order. For example, with text-shadow, you must do "color" followed by "values" or vice versa. Also, you must have both to specify color (you can't do `text-shadow: red`) for example.  
  
With flex-flow though, you can do any combination including omitting either, i.e
    
       ~ flexFlow wrap  
       ~ flexFlow row  
       ~ flexFlow wrap row  
       ~ flexFlow row wrap 
  
[...]

PR #82, opened 2016-02-18
add display: flex

jgillich:
I'm not sure how to name this one, since `flex` is obviously taken.

ref #13


lukewestby:
this is the case for a couple other property keys/values as well, like `top` and `bottom`. can you take a look at https://github.com/rtfeldman/elm-css/blob/master/src/Css.elm#L521-L529 as a starting point for how this was handled for those?

jgillich:
I've looked into it, but this is all a bit much for me right now; as mentioned in the original issue, I'm just getting started with Elm. Will revisit this in a couple of days/weeks.

lukewestby:
sounds good! anyone you can find on this repo or in the elm slack channel or elm-discuss mailing list would be happy to help out as you're working on it!

rtfeldman:
I added `displayFlex` to implement this. Thanks!

PR #85, opened 2016-03-05
Add font-feature-settings

lukewestby:
I think technically feature tags should be implemented as a union type, but there are a looot of them (https://www.microsoft.com/typography/otspec/featurelist.htm) so for now I left it as a string to get your opinion about that implementation detail.


src/Css.elm
```diff
@@ -3236,6 +3242,58 @@ slashedZero =
   }
 
 
+{- FEATURE TAG VALUES -}
+
+
+{-| Alias for `on` value of [`font-feature-settings`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-feature-settings#Formal_syntax)
+-}
+on : Int
+on =
```
eeue56: So one of the things we found was that if you were trying to have styles alongside your HTML, using names like `on` would collide with things from elm-html. So this would likely become `Css.on` in most real code. I wonder if a longer, more qualified name, would be better, since then it won't collide? Especially for something that is rarely actually going to be used, because this will mean that I will have to use `Html.Events.on` too

lukewestby: Ah! I completely forgot about `Html.Events.on`. So I was thinking, this seems to be an advanced feature of CSS anyway so its users probably know that `off` corresponds to 0 and `on` corresponds to 1. Do you think we can just exclude `on` and `off`?

eeue56: The spec defines it as

```elm
normal | <feature-tag-value>#
where 
<feature-tag-value> = <string> [ <integer> | on | off ]?
```

So I think you need something similiar to on/off

lukewestby: It also says "The two keywords on and off are synonyms for 1 and 0 respectively" so I thought it might be okay just to leave them out. If y'all think we should include them how about `featureOn` and `featureOff`?

eeue56: No strong opinion other than it shouldn't be _just_ `on`, `off`

lukewestby:
Sounds good. I renamed the to `featureOn` and `featureOff`, and also realized I forgot to cover `normal`, so that's in there now as well.

Issue #86, opened 2016-03-05
Node version support


lukewestby:
Should the command line tool be supported for node < v4? The current implementation uses template strings which aren't implemented in < v4 so CI is failing for those runtimes.

eeue56:
I'm torn. Happy to remove template strings if we want to support old node, but I really think that supporting 4+ is fine. There's a few posts talking about how 0.12 still has wide usage, but I don't think those companies are the ones using Elm or interested in using elm-css.

lukewestby:
I agree. I think 0.12 having wide usage is more a concern for server-oriented modules than it is for tooling-oriented ones.

eeue56:
What do you think @rtfeldman ?

rtfeldman:
Yeah, I think 4+ is fine. :smile:

PR #88, opened 2016-03-06
add node version requirement

lukewestby:
For #86 :smile:


eeue56:
:+1: probably worth moving everything that can be to template strings too, like this guy - https://github.com/rtfeldman/elm-css/blob/master/index.js#L125

Don't think that has to be in this PR though

PR #102, opened 2016-04-01
make hex function can preprocess "#2B414D" like color String 

raiscui:
(have "#……") ,so some codeIDE can colored on them

image

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -852,14 +852,21 @@ hsla hue saturation lightness alpha =
 hex : String -> Color
 hex str =
   -- TODO
-  { value = "#" ++ str
-  , color = Compatible
-  , red = 0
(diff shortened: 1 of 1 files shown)


eeue56:
You need to update the comment to explain this new behaviour

raiscui:
sorry I'm not have many experience for github,  I changed hex function make it can accept / recognition the str have char:"#" or not, because the color string with "#" will can colored string by ATOM or other IDE ,that's very helpful "Coder/Designer experience" of css work

eeue56:
Got it. I meant you need to document the change to the function, by changing the comment for the function to say how it behaves now if you pass it a hexvalue with a hash

rtfeldman:
Good idea, thanks!

Issue #32, opened 2015-12-24
Validity checks


rtfeldman:
This is not MVP, but at some point I'd like to have `elm-css` do some additional checks for invalid values (perhaps once the Declarations have been assembled, or perhaps sooner...waiting for Declarations arguably allows extra context and thus nicer error messages, but requires more regexp work).

Some checks that sound good:
- [x] Run a regex (`regex "^([a-fA-F]{3}|[a-fA-F]{6})$"` for hex colors) to confirm they're valid color values.
- [ ] Warn if there are any selectors defined with no properties; they won't do anything and should be deleted.
- [ ] Warn if the same property is declared multiple times within the same selector. All but the last declaration should be deleted.
- [ ] Warn if the same selector is declared multiple times within the same stylesheet. There are theoretically reasons it might be reasonable to support this, but more likely than not, this will just be error-prone. The Anti Footgun League says: complain about it.
- [ ] Warn if any MultiSelectors include multiple [Type Selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Type_selectors); these will for sure emit busted CSS.
- [ ] Warn if any class, id, or animation names are not [valid CSS identifiers](http://stackoverflow.com/questions/448981/which-characters-are-valid-in-css-class-names-selectors) - possible if you use strings instead of union types.
- [ ] Verify that [text-align](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align) and [text-align-last](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align-last) disallow the appropriate properties that the other doesn't support. This can't be fixed with type checking because of `left`, `right`, `top`, and `bottom` being overloaded as both keys and values.

rtfeldman:
Now that `~` just resolves mixins, I think the properties themselves can handle these - and with context. Just need to have `InvalidStyle` build up an error message context instead of being a no-op so often, and have stuff like `prop1` incorporate the notion of adding to the error context.

Then we can get fail fast, nice error messages, etc.

tolgap:
@rtfeldman While working on #105 I noticed that `propN` methods are sometimes used with multiple `type aliases` that do not have a notion of `warnings` attached to them, like `textDecoration3`:

```{elm}
textDecoration3 : TextDecorationLine compatibleA -> TextDecorationStyle compatibleB -> ColorValue compatibleC -> Mixin
textDecoration3 =
  prop3 "text-decoration"
```

[...]

PR #105, opened 2016-04-07
Add hex color validations

tolgap:
I've been looking to get into Elm, and this project seems a great effort to be able to reduce the errors in writing CSS in combination with Elm.

The first item in issue #32 , validation checks are asked for hex values. I have implemented these.

Any feedback is very much welcome as I'm still learning!

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -4189,8 +4189,8 @@ backgroundColor =
 
 {-| -}
 color : ColorValue compatible -> Mixin
-color =
-  prop1 "color"
+color c =
+  propertyWithWarnings c.warnings "color" c.value
 
 
 {-| -}
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -863,14 +863,14 @@ hex str =
         "#" ++ str
     warnings =
       if contains (regex "^#([a-fA-F0-9]{8}|[a-fA-F0-9]{6}|[a-fA-F0-9]{4}|[a-fA-F0-9]{3})$") value then
+        []
+      else
         [ String.join " "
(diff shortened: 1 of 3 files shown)


src/Css.elm
```diff
@@ -854,21 +855,25 @@ tools which express these as e.g. `#abcdef0`, etc.
...
-  -- TODO emit warnings for invalid hex chars
   let
     value =
       if String.slice 0 1 str == "#" then
         str
       else
         "#" ++ str
+    warnings =
+      if contains (regex "^#([a-fA-F0-9]{6}|[a-fA-F0-9]{3})$") value then
+        [ "Hexadecimal color values can only contain three or six characters of '0-9' and/or 'A-F'. " ++ value ++ " is not valid." ]
```
eeue56: Generally prefer to join strings with like

```
String.join " "
  [ "hello"
  , value
  , "dog"
  ]
```

tolgap: @eeue56 I wanted to perform this change, but I noticed that all string concatenation in the file happens with `(++)`. Should I change it anyway?

rtfeldman: current CSS4 draft [supports lengths of 4 and 8](https://drafts.csswg.org/css-color/#hex-notation) (for alpha); I'm inclined to think allowing those is okay. 😄

rtfeldman: Also, 👍  for `String.join` here

tolgap:
@rtfeldman I performed the necessary changes.

I could squash the commits if necessary, but GitHub has built in squash merging now. Or do you want to keep the commits as is?

rtfeldman:
Love it, thanks @tolgap! 😻

PR #107, opened 2016-04-13
A couple of fixes on the Tutorial

devnacho:
Hey @rtfeldman, 

First of all, thanks for elm-css... I really like css-modules but this takes it even further! :grinning: 

I was trying to learn how to do media queries with the tutorial and notice that there are some things that might be outdated. 
Those things are:
-  `with hover` for `hover`
-  `with media "print"` to `withMedia [ print ]`

What I don't know how to fix, and no longer work is: 
- `with ((.) Large) `
-  `with ((#) SignUp)`

Should it be `withClass ((.) Large)` ?

Cheers,

Nacho

### DIFF
--- a/Tutorial.md
+++ b/Tutorial.md
@@ -144,13 +144,13 @@ stylesheet { name = "homepage" }
   , a
       [ textDecoration none
 
-      , with hover
+      , hover
           [ textDecoration underline ]
       ]
   ]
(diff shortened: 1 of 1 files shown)


rtfeldman:
Awesome, thanks @devnacho! 🎉

PR #108, opened 2016-04-15
Fix broken testsuite

roman:


rtfeldman:
Thanks so much @roman!

PR #110, opened 2016-04-18
Flesh out README with all the code needed to make this work

jessitron:
The examples in the readme were not complete. They didn't give me enough information to make elm-css work for me.

I think this adds enough. (addresses issue 87 among other things)

In addition, I added a directory containing the code for the README to prove that it compiles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Here's an example:
 
 ```elm
 css =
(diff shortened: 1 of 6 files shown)


eeue56:
Seems good to me! Thanks a lot!

PR #113, opened 2016-04-20
remove quotes on media queries

devnacho:
Hey @rtfeldman, 

I just noticed that the output of  media queries had a syntax error(extra quotes). The output of media queries now is:
```css
@media "print"{
}
// or 
@media "screen and (max-width: 400px)" {
}
```

and the correct syntax according to the [spec](https://developer.mozilla.org/es/docs/CSS/Media_queries) is this one:
```css
@media print{
}
// or 
@media screen and (max-width: 400px) {
}
```

I changed it, updated the docs and the tests.

Let me know if code looks good and sorry for not catching this error in the previous PR. 

Cheers,

### DIFF
--- a/Tutorial.md
+++ b/Tutorial.md
@@ -270,6 +270,10 @@ stylesheet { name = "homepage" }
           [ width (px 1280) ]
       ]
 
+  , mediaQuery "screen and ( max-width: 600px )"
+      [ body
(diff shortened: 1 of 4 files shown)


rtfeldman:
No worries! Good call.

rtfeldman:
Published as 2.1.1.

PR #122, opened 2016-05-15
Correcting some docstrings that refer to the obsolete 'n' function.

krisajenkins:
This was replaced with 'int' in commit: 660c8b29ec292cbfb7c62ea4d4037e097be00756


rtfeldman:
Ah, nice catch! Thanks! 💯

PR #123, opened 2016-05-15
Adding h5 and h6 selectors.

krisajenkins:
I'm trying to port http://getskeleton.com/ to Elm, as a field-test. This is one of the missing pieces I've come up against.


rtfeldman:
Seems reasonable, thanks @krisajenkins!

PR #124, opened 2016-05-15
Fix elm-css cli for Elm version 0.17.

### LINKED ISSUE #120: Ports error syntax
Hey @rtfeldman @eeue56,

Thanks for releasing the new update for 0.17 👏🏼

I'm running elm-css
`elm-css src/elm/Stylesheets.elm -o src/styles`

and I'm getting this Syntax Error
```
-- SYNTAX PROBLEM -------------------------------------- src/elm/Stylesheets.elm

I ran into something unexpected when parsing your code!

11| port files =
               ^
I am looking for one of the following things:

    the "has type" symbol ':'
    whitespace
```

Is it possible that the following no longer works with 0.17 syntax of [defining ports](http://guide.elm-lang.org/interop/javascript.html)?

```
port files : CssFileStructure
port files =
    toFileStructure [ ( "homepage.css", compile Homepage.css ) ]
```

If it's of any help this is my [Stylesheets port module](https://github.com/devnacho/front-vientus/blob/master/src/elm/Stylesheets.elm) and from what I see it is written exactly like the example.

Thanks, 

Ignacio

jwoudenberg:
Due to the changes in the port system on both the JavaScript and Elm
side the cli application was no longer able to extract the generated
CSS from the Elm appliation.

This commit fixes the issue in the CLI and updates the example code so
the two work together again.

Resolves #120.

Some comments:

- I had to include elm-css-util in the examples elm-package.json as well, or I'd get an error that Css.Helpers could not be found.
- The one way I could figure out to publish a command now was to create a complete Program and send the command as part of the Program's init step. The one Program currently available is `Html.App`, so that's the one I used. Does someone know an easier way?
- The JavaScript generated by Elm now self-exports `Elm` on `module.exports`.

### DIFF
--- a/examples/src/Stylesheets.elm
+++ b/examples/src/Stylesheets.elm
@@ -2,8 +2,23 @@ port module Stylesheets exposing (..)
 
 import Css.File exposing (..)
 import HomepageCss as Homepage
+import Html exposing (div)
+import Html.App as Html
 
 
-port files : CssFileStructure
-port files =
+port files : CssFileStructure -> Cmd msg
+
+
+cssFiles : CssFileStructure
+cssFiles =
     toFileStructure [ ( "homepage.css", compile Homepage.css ) ]
+
+
+main : Program Never
+main =
+    Html.program
+        { init = ( (), files cssFiles )
+        , view = \_ -> (div [] [])
+        , update = \_ _ -> ( (), Cmd.none )
+        , subscriptions = \_ -> Sub.none
+        }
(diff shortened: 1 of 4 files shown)


examples/src/Stylesheets.elm
```diff
@@ -2,8 +2,23 @@ port module Stylesheets exposing (..)
 
 import Css.File exposing (..)
 import HomepageCss as Homepage
+import Html exposing (div)
+import Html.App as Html
 
 
-port files : CssFileStructure
-port files =
+port files : CssFileStructure -> Cmd msg
```
stoeffel: this also needs to be updated in the readme

rtfeldman:
Love it! 😻  Thank you @jwoudenberg

PR #132, opened 2016-06-08
Readme/example: Change ID from list to single item

stkb:
In the readme example, shouldn't the `id` function be taking a single identifier for the css ID instead of a list? I understand that it still works as a list, but it seems more normal to have it as a single identifier, and it may otherwise cause some head-scratching for newcomers (as it did with me).

This PR contains the needed changes if you agree.


rtfeldman:
I agree, thanks @stkb!

PR #138, opened 2016-07-05
Remove space before pseudo element in compiled CSS

bardt:
Right now compiled CSS with pseudo elements looks like `.selector ::before {}`. This space before the pseudo element is misinterpreted by browsers. This whole rule will apply `::before` to `.selector`'s children.


rtfeldman:
Nice, thank you!

rtfeldman:
Added some tests and published as 3.1.2! Thanks again @bardt! 😸

PR #141, opened 2016-07-18
Update example dependencies to run on Elm 0.17.1

fredcy:
I was unable to build the examples with Elm 0.17.1 until I relaxed the elm-version constraint.

While doing that I also relaxed the core constraint to allow the new 4.0.x releases.


rtfeldman:
Thanks @fredcy! :metal:

PR #144, opened 2016-07-22
Cursor property

tolgap:
This implements the cursor property.

I refrained from defining the `cursor: default` value as a function called `default`. Instead I opted for a function called `cursorDefault`. If necessary, I can rename it to `default`.


rtfeldman:
Curious - why not call it `default`?

rtfeldman:
Also, test failure looks legit 😄 

screen shot 2016-07-22 at 9 15 45 am

tolgap:
I had this weird feeling that having a function called `default` just for a cursor value would be "overkill" 😆. Anyway, I named it default, and I fixed the test failure.

PS: my local test suite did not fail with that error, any idea why?

rtfeldman:
> PS: my local test suite did not fail with that error, any idea why?

Stale cache is the only thing that comes to mind. Try deleting `elm-stuff/build-artifacts` and re-running?

tolgap:
@rtfeldman nuking `build-artifacts` helped me find the test errors, thanks for the tip.

rtfeldman:
Lovely! Thanks for this @tolgap! 😻

rtfeldman:
Published as 4.0.0!

PR #145, opened 2016-07-22
Add support for compiling many stylesheets

scottcorgan:
I'm open to rename the function, but I wasn't sure what the convention was for describing an _n_ number of arguments. Maybe it's something like `compileN`?

### DIFF
--- a/examples/src/Stylesheets.elm
+++ b/examples/src/Stylesheets.elm
@@ -11,7 +11,7 @@ port files : CssFileStructure -> Cmd msg
 
 cssFiles : CssFileStructure
 cssFiles =
-    toFileStructure [ ( "homepage.css", compile Homepage.css ) ]
+    toFileStructure [ ( "homepage.css", compile [ Homepage.css ] ) ]
 
 
 main : Program Never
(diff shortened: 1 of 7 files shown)


rtfeldman:
Cool! I feel like we should just replace `compile` with this. Doesn't really seem like there's a need for both functions, and this one is more useful.

I'd avoid `foldr` though, so we don't make so many intermediate data structures - something like:

```elm
compile : List Preprocess.Stylesheet -> { warnings: List String, css: String }
compile styles =
    let
        results =
            List.map compile styles
    in
        { warnings = List.concatMap .warnings results
        , css = List.join "\n\n" (List.map .css results)
        }
```

scottcorgan:
Fair enough. I didn't think that memory/extra data structures mattered as this isn't a feature with too much of a performance concern. I'll try some recursion though! Thanks!

scottcorgan:
Amended to last commit

rtfeldman:
Fantastic, thanks! I'm gonna merge this now, but unless it's urgently needed, I'd rather wait to publish until there's another breaking change (which actually I suspect `float` will be).

scottcorgan:
Thanks! No rush, I'm just using a custom function until it gets merged.

Issue #140, opened 2016-07-12
`each` with pseudo classes only apply to last selector


matsrietdijk:
```
each [input, selector "textarea"]
    [ focus
        [ borderColor (hex "#000000")
        ]
    ]
```
Currently results in:
```
input, textarea:focus {
    border-color: #000000;
}
```
Expected was:
```
input:focus, textarea:focus {
    border-color: #000000;
}
```

rtfeldman:
Thanks @matsrietdijk!

This is still reproducible in 4.0.1.

I wrote a failing test to reproduce it on a branch: edf83471b9aaa11c95c86c13f87d1086927bde98 - test output is:

![screen shot 2016-07-22 at 9 56 43 am](https://cloud.githubusercontent.com/assets/1094080/17064681/a0d45c0e-4ff2-11e6-8058-a1a3e8a70cd6.png)

rtfeldman:
Fixed in https://github.com/rtfeldman/elm-css/pull/146 by @tolgap, who is on FIRE with the contributions! 🎆

PR #146, opened 2016-07-24
Fix pseudo elements and classes not resolving when chained with `each` 

tolgap:
This attempts to fix issue #140 .

While working on the fix, [I noticed that `mapLast` is used instead of `List.map`](https://github.com/rtfeldman/elm-css/blob/7c58295a3b21efc6b7a6904e64df36bf36a58529/src/Css/Structure.elm#L259) when resolving chained pseudo classes. @rtfeldman even though I changed it to `List.map`, I can't figure out if it's a correct assumption that `mapLast` was at fault here. Could I have regressed something by changing it to `List.map`? I've thought about it, and couldn't find a scenario that would fail tests.

Also, while working on it, I realised the same issue applied to pseudo *elements*, so I fixed that as well, and introduced a nice refactor to remove some duplication.

Thoughts and concerns? :)

### DIFF
--- a/src/Css/Preprocess/Resolve.elm
+++ b/src/Css/Preprocess/Resolve.elm
@@ -279,7 +279,7 @@ applyMixins mixins declarations =
             applyNestedMixinsToLast
                 nestedMixins
                 rest
-                (Structure.appendToLastSelector selector)
+                (Structure.appendRepeatableToLastSelector selector)
                 declarations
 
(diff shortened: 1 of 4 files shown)


rtfeldman:
Wow, great work @tolgap! I also dig the refactor. 😻

rtfeldman:
Published as 4.0.2!

PR #147, opened 2016-07-27
add program function

lukewestby:
This adds a function to set up the `Program` you need in order to interface with the CLI. Totally cool if y'all think this is unnecessary, I just figured this set up is probably going to look exactly the same every time so it might as well be included.

### DIFF
--- a/elm-package.json
+++ b/elm-package.json
@@ -14,6 +14,7 @@
     ],
     "dependencies": {
         "elm-lang/core": "4.0.0 <= v < 5.0.0",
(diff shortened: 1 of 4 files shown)


rtfeldman:
I really want to avoid `elm-css` itself having a dependency on `virtual-dom` or `html` based on past experiences with old `elm-test` where version bounds caused conflicts with application code dependencies.

I like the idea, but I think it belongs in a different package. (Not sure exactly where though!)

lukewestby:
sounds good! always good to learn from past experience.

PR #149, opened 2016-07-29
Updated Partial Implementation of Post-1.0.0 keywords

colvardrm:
I just updated the rest of my project to match the current master branch, then merged the changes I did to Css.elm with the updated project. It looks like the conflicts were just due to my version being really outdated.


src/Css.elm
```diff
@@ -2324,6 +2404,16 @@ prop3 key argA argB argC =
...
+    
+    
+{-| Sets ['float'](https://developer.mozilla.org/en-US/docs/Web/CSS/float)
+floatAlign : FloatAlign compatible -> Mixin
+
+    floatAlign right
+-}
+floatAlign : (ExplicitLength IncompatibleUnits -> Mixin) -> Mixin
+floatAlign fn =
+    getOverloadedProperty "floatAlign" "float" (fn lengthForOverloadedProperty)
```
rtfeldman: I'd say we should call this `float` and rename the existing `float` to `num`.

(In retrospect, I think calling the existing function `float` was a mistake.)

Thoughts @colvardrm?

colvardrm:
That sounds like a reasonable change. I can try to implement it.

src/Css.elm
rtfeldman: @colvardrm I changed `float` to `num` on `master` so you're all set to rename! Also I split the exports line into multiple lines so it should be easier to avoid merge conflicts this time. 😅

Issue #148, opened 2016-07-28
Does `color inherit` not work?


tomekwi:
Hey there,

looking at https://github.com/rtfeldman/elm-css/issues/98 I see that it’s intended to work. But then again, here’s the error I get when I try to use `color inherit` in my code:

<details><summary>```The argument to function `color` is causing a mismatch.```</summary>
```elm
136|       color inherit
                 ^^^^^^^
Function `color` is expecting the argument to be:

    { a | ..., color : ..., warnings : ... }

But it is:

    { ...
    , alignItems : ...
    , all : ...
    , boxSizing : ...
    , display : ...
    , flexBasis : ...
    , flexDirection : ...
    , flexDirectionOrWrap : ...
    , flexWrap : ...
    , fontFamily : ...
    , fontSize : ...
    , fontStyle : ...
    , fontVariant : ...
    , fontWeight : ...
    , length : ...
    , lengthOrAuto : ...
    , lengthOrMinMaxDimension : ...
    , lengthOrNone : ...
    , lengthOrNoneOrMinMaxDimension : ...
    , lengthOrNumber : ...
    , lengthOrNumberOrAutoOrNoneOrContent : ...
    , none : ...
    , number : ...
    , numericValue : ...
    , overflow : ...
    , textDecorationLine : ...
    , textDecorationStyle : ...
    , textIndent : ...
    , textRendering : ...
    , unitLabel : ...
    , units : ...
    }
```
</details>

If you need any more info, I’ll be glad to help.

rtfeldman:
Fixed in https://github.com/rtfeldman/elm-css/pull/150

PR #150, opened 2016-07-29
Allow color to be inherit, unset and initial

tolgap:
This aims to fix issue #148 . [The CSS spec](https://developer.mozilla.org/nl/docs/Web/CSS/color) allows color to assume the following global values:

```css
/* Global values */
color: inherit;
color: initial;
color: unset;
```


rtfeldman:
Lovely! 😍 

Thanks @tolgap!

PR #151, opened 2016-07-31
outline, outline-width, outline-style, outline-color and outline-offset properties

tolgap:
This implements the `outline` property (and co).

Even though the CSS spec of `outline` does not mention this (for some weird reason), but browsers accept `outline: 0` as well. So I made it also accept `outline zero`. It is also properly configured to accept the global values `initial`, `inherit` and `unset`.


rtfeldman:
Nice, thanks @tolgap! 🎉

Issue #158, opened 2016-08-27
How are fonts supposed to be set?


SandiDusic:
[This tutorial](https://github.com/rtfeldman/elm-css/blob/5b10188fa0bb221a9b6de1a8b1dba46472a22ef7/Tutorial.md) has the following line of code:
```[ fontFamily [ "Georga", "serif" ]```

But when I try something similar in my code, I get the following error:
```
The argument to function `fontFamily` is causing a mismatch.
Function `fontFamily` is expecting the argument to be:
    { a | fontFamily : Css.Compatible, value : String }
But it is:
    List String
```

The compiler accepts `fontFamily sansSerif` (or serif), but that's not exactly a plethora of choices.

tolgap:
That should be `fontFamilies` (so plural). Will need to fix that example! If you wouldn't mind, you could apply the fix to that example and open a pull request :)

SandiDusic:
Sure, I'll do so tomorrow. Thanks for the clarification!

SandiDusic:
Here: https://github.com/rtfeldman/elm-css/pull/159

PR #159, opened 2016-08-28
Improve font documentation and fix tutorial error

SandiDusic:
As per the suggestion I received in https://github.com/rtfeldman/elm-css/issues/158, this PR fixes a tiny error in the tutorial. Since I came across that error by searching for a way to set fonts by their name (like "Helvetica" instead of sansSerif or whatever), I took the liberty of adding a bit of documentation to `fontFamily` pointing to `fontFamilies`, since that's the first place I went looking. After all, the CSS property is called `font-family`, and there is no `font-families` one.

### DIFF
--- a/Tutorial.md
+++ b/Tutorial.md
@@ -39,7 +39,7 @@ stylesheet { name = "homepage" }
       ]
 
   , (.) MenuItem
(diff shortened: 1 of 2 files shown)


rtfeldman:
Very nice! Thanks @SandiDusic!

PR #164, opened 2016-09-27
Add table display styles

miniBill:


miniBill:
Better?

rtfeldman:
Excellent, thank you @miniBill!

rosenk:
table, tableRow, tableCell are not exposed from the module and I can't use them. Or I'm missing something obvious?

tolgap:
You are correct @rosenk and tests would have picked those up if they were present in this PR. Perhaps you could do a fix-PR for this included with tests? :)

rosenk:
I don't beleave in tests. The compiler should have a warning if a function is neither used nor exposed.

rtfeldman:
@rosenk It presumably will at some point in the future. However, that would not give us any confidence that the functions are actually emitting what we expect.

Tests are good.

rosenk:
I just finished my very first real elm project and I'm very grateful for the incredible work you are doing. I centrally want to contribute back but I'm very new to all of it. I don't know how to write such tests. Any links to point me in the right direction?

PR #165, opened 2016-09-28
Remove unneeded index.html and tests/homepage.css files

tolgap:
I noticed these two files in the repo, and being generated files, they _should_ be ignored. Especially the `index.html` still containing Elm 0.16 javascript and inheritely causing the following report:

screen shot 2016-09-28 at 21 10 11


rtfeldman:
Nice, thanks @tolgap!

PR #168, opened 2016-10-01
Add support for list-style, list-style-type, list-style-position

vipentti:
Add basic support for `list-style`, `list-style-type` and `list-style-position`. 

* [`list-style-position`](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-position) implements the full spec from MDN.

* [`list-style-type`](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type) currently only implements some of the predefined styles from [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type#Values) that have not been marked experimental. 
* [`list-style-image`](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-image) is currently _not_ implemented. This requires a good way of defining URLs which could also be used for [`background-image`](https://developer.mozilla.org/en/docs/Web/CSS/background-image). 

* [`list-style`](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style) shorthand currently supports `list-style-type` and `list-style-position`, however, there are no checks in place so you could technically pass the same type of property several times. I'm not sure if there is a way to check against this (or if it is even needed).


rtfeldman:
Looks great @vipentti! Thanks! 😻

PR #171, opened 2016-10-04
Fix: rgba was not setting alpha in Color

### LINKED ISSUE #166: rgba sets { color | alpha = 1 }
Wasn't sure if this was a bug or there for a reason, but https://github.com/rtfeldman/elm-css/blob/master/src/Css.elm#L823 always sets alpha to 1 regardless of the input parameter.

martindemello:
Fixes #166


rtfeldman:
Thanks @martindemello!

PR #173, opened 2016-10-05
Box shadow

rodrigo-morais:
Adding the box-shadow property in Css properties.


rtfeldman:
Fantastic! Thank you @rodrigo-morais!

PR #176, opened 2016-10-12
Fix line-height not accepting numbers

### LINKED ISSUE #175: lineHeight with Number
Spec of line-height property is
```
line-height: normal|number|length|initial|inherit;
```

However, in elm-css,
```
lineHeight : Length compatible units -> Mixin
```

Because input of lineHeight only allows `length`, it doesn't allow `number`.

For only one element, `number` makes same result with `percent`, which included in `length`, but for inherited elements, `number` properties aren't computed to fixed length where `percent` are. Because of this difference, `number` properties are preferred in many case (according to https://developer.mozilla.org/en/docs/Web/CSS/line-height).

Is there any way to use `number` in `lineHeight`? Did I misread elm-css? If not, could you please concern about this issue?

tolgap:
Fixes #175. I ran into this myself in the past, might as well do a fix as others seem to be running into too 😆

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -5652,7 +5652,7 @@ withMedia =
 
     lineHeight (px 10)
 -}
-lineHeight : Length compatible units -> Mixin
+lineHeight : LengthOrNumber compatible -> Mixin
 lineHeight =
     prop1 "line-height"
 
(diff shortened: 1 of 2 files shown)


rtfeldman:
Ha, thank you @tolgap! 😄

PR #177, opened 2016-10-13
Adding support for the  justify-content flex-box property and the spa…

tommcgurl:
This Adds support for flex-box's `justify-content` property as well as the `space-around` and `space-between` values. 

More can be read about this property [here](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content)

You can read about the `space-around` and `space-between` values [here](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content#Values)

Thanks!

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -369,11 +369,14 @@ module Css
         , order
         , alignItems
         , alignSelf
+        , justifyContent
         , content
         , wrapReverse
         , wrap
         , flexStart
         , flexEnd
+        , spaceAround
+        , spaceBetween
         , stretch
         , row
         , rowReverse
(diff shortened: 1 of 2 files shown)


src/Css.elm
```diff
@@ -3092,6 +3104,21 @@ alignSelf fn =
...
+  use this workaround:
+
+  property "justify-content" "auto"
+
+  If this is annoying, please file an issue, so adding support for "auto"
+  can be prioritized!
+-}
+justifyContent : (ExplicitLength IncompatibleUnits -> Mixin) -> Mixin
+justifyContent fn =
+    getOverloadedProperty "justifyContent" "justify-content" (fn lengthForOverloadedProperty)
```
rtfeldman: Why not ship with `auto`? 🤔

tommcgurl: Good call, I'll make this change. I was under the impression it was not supported across the board and thought maybe it would be worth adding everywhere at once. Now that I'm looking it seems like it's easy enough to just add it here. 👍

tommcgurl: Ah so as it turns out `auto` is not an option for `justify-content`. It is only a value for `align-self`. I've removed the comment regarding the `auto` property.

rtfeldman:
Thanks @tommcgurl! If you can resolve the merge conflicts, this is ready to merge. 🤘

tolgap:
@tommcgurl also, your last commit has added 3 unnecessary files. Can you remove those?

tommcgurl:
Okay all set 👍

rtfeldman:
Lovely, thank you @tommcgurl!

tommcgurl:
No problem thank YOU for this awesome project 😊👍

Issue #186, opened 2016-11-03
z-Index


amitu:
How to set zIndex CSS property?

tolgap:
Z-index is not in elm-css yet. I use the following function:

```elm
zIndex : Int -> Mixin
zIndex i =
    property "z-index" <| toString i
```

If there is a property not present in elm-css, you can use `property` to define it yourself.

amitu:
@tolgap Thanks :-)

Would be useful to have that bit in README.md (along with how to create a non covered pseudo selector say).

tolgap:
@amitu you're right. Most opened issues or questions on Slack are about missing properties that can be resolved using `property` function.

rtfeldman:
@amitu @tolgap that one is super easy to add if anyone wants to make a PR 😉

lukewestby:
#191

it's been a while so i took a stab at it. it turns out that if we want to cover all the possible values for z-index it's not so simple, but covering just ints is fine

rtfeldman:
Added in https://github.com/rtfeldman/elm-css/pull/195

Issue #174, opened 2016-10-06
Add content to pseudo element


rodrigo-morais:
In pseudo elements as `:before` and / or `:after` is possible to use `content` property.
Today we have a `content` property linked with `flex`. I'd like add a `content` to `:before` and `:after`.

What is the better design for that?

We could reuse the same content and verify the parameter. If the parameter is a `String` then we use it as a property to pseudo element, but if the parameter is a `LengthOrNumberOrAutoOrNoneOrContent` than we use it as property to `flex`.

Another option is to create a new property with another name as `contentPseudoElement`.

What is the best solution?
Have any other?

scottcorgan:
For now, you can use `property "content" "''"`

rtfeldman:
https://github.com/rtfeldman/elm-css/issues/432 would make it possible to introduce a `content` property without namespace collisions!

PR #187, opened 2016-11-10
Explain how missing properties can be defined

tolgap:
Most opened issues or questions on Slack are about missing properties that can be resolved using property function.

Examples: #186 #174 and some in Slack

So including this inside the README might help.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,26 @@ view =
 
 ```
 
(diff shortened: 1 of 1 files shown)


rtfeldman:
Great idea! 💯 

Thanks @tolgap!

PR #191, opened 2016-11-17
add simple zIndex

lukewestby:
This is the most obvious way to do this, but excludes the `auto` value.

Including `auto` will be an interesting types problem. `z-index` is only allowed to have integers or `auto`, but the `auto` value currently has a much broader type. We could make another `IntOrAuto` type but then we need a function that converts `Int` to `IntOrAuto`, which might be weird because `int` already exists and is also too broad.

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -98,6 +98,7 @@ module Css
         , height
         , minHeight
         , maxHeight
+        , zIndex
         , padding
         , padding2
         , padding3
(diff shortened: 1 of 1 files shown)


rtfeldman:
Hm...I feel like in order to be consistent with how other things have done it, we should go with `zIndex (num 5)` so you can have `zIndex auto` and `zIndex inherit` and the like.

Thoughts @lukewestby?

lukewestby:
Do you think it's alright that it would let someone do `zIndex (num 1.2)` and end up with broken CSS?

rtfeldman:
@lukewestby ha, good point! I should have said [`int`](http://package.elm-lang.org/packages/rtfeldman/elm-css/7.0.0/Css#int) instead.

lukewestby:
But is there a way to require that someone only uses `int`, `auto`, or any of the `BasicProperty`s in the system?

rtfeldman:
@lukewestby got it working! https://github.com/rtfeldman/elm-css/pull/195

rtfeldman:
Takes a major version bump though, so I'll hold off on it for a bit.

rtfeldman:
Closing in favor of #195

PR #193, opened 2016-11-18
Fix tests after 0.18 upgrade

tolgap:
Started debugging the test issue of failing tests after the 0.18 upgrade. And found an example issue in the process.


rtfeldman:
Nice, thank you @tolgap!

PR #201, opened 2016-12-09
💥 Rename (.) to class and (#) to id 💥

rtfeldman:
💥 Big breaking change! 💥

This has been a long time coming, but it seems like the right choice:

Replace `(.)` with `class` and `(#)` with `id`.

Even though these are nicer names, I've avoided doing this because I wanted to avoid namespace clashes in the case where you're defining styles in the same file as some view code, and you have `import Css exposing (..)` and `import Html.Attributes exposing (..)` in the same file.

However, what I realized is that:

1. This already comes up with stuff like `src`, and if you do `import Html.Attributes as Attr exposing (..)` followed by `Attr.src` to disambiguate and resolve the compiler error, it's not a huge deal.
2. In practice, if you're writing views and styles in the same file, you're probably shadowing `class` anyway using a namespace helper (e.g. `{ class, classList, id } = ...`) so the only downside would be that you'd have to say `Css.class` instead of `(.)` - which is more verbose, but reads better. Seems like a win overall.

My default plan is to merge this, but I want to leave it open for a bit in case someone brings up a compelling point I missed. It's a big change, after all!

### DIFF
--- a/examples/src/HomepageCss.elm
+++ b/examples/src/HomepageCss.elm
@@ -18,17 +18,17 @@ css =
             [ display inlineBlock
             , paddingBottom (px 12)
             ]
-        , (.) NavLink
+        , class NavLink
             [ margin (px 12)
             , color (rgb 255 255 255)
             ]
-        , (#) ReactiveLogo
+        , id ReactiveLogo
             [ display inlineBlock
             , marginLeft (px 150)
             , marginRight (px 80)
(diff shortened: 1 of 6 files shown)


rtfeldman:
Well, this has been open for over a month and the current reactions are:

screen shot 2017-01-17 at 10 48 02 pm

Guess this is ready to merge!

Issue #183, opened 2016-11-01
display: table


butchcom:
I think that there's no support for `display: table` and derivatives (or I'm just not able to find them). I'd really appreciate it, if that could be added.

tolgap:
@butchcom display: table issue is also mentioned here: https://github.com/rtfeldman/elm-css/issues/111

butchcom:
Ok, I didn't notice that issue.

rtfeldman:
@butchcom @tolgap would be happy to accept a PR to add it!

PR #205, opened 2016-12-31
Add `display:table` and its derivatives

folkertdev:
adressing #183 

The basic table properties were implemented, but not exposed. This
PR adds

* table
* inlineTable
* tableCell
* tableRow
* tableColumn
* tableCaption
* tableRowGroup
* tableColumnGroup
* tableHeaderGroup
* tableFooterGroup

and their associated tests

further notes:
* `table` is already a selector (the table html
element) and thus `display table` will have to become `display Css.table` in
most actual use cases where elm-css is imported as `exposing (..)`. A specific `displayTable` property (analogous todisplayFlex) is a possible fix.
* In the comments, I've changed the text in the link to the css value instead of the elm-css value (`inline-table` instead of `inlineTable`). This is consistent with other links to the MDN documentation. 
* Is the ordering of the properties alright?

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -251,6 +251,16 @@ module Css
         , block
         , inlineBlock
         , inline
+        , table
+        , inlineTable
+        , tableCell
+        , tableRow
+        , tableColumn
+        , tableCaption
+        , tableRowGroup
+        , tableColumnGroup
+        , tableHeaderGroup
+        , tableFooterGroup
         , listItem
         , inlineListItem
         , none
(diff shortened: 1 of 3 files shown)


src/Css.elm
```diff
@@ -3666,7 +3687,7 @@ tableRow =
     }
 
 
-{-| Sets the display style to [`tableCell`](https://developer.mozilla.org/en-US/docs/Web/CSS/display#Values)
+{-| Sets the display style to [`table-cell`](https://developer.mozilla.org/en-US/docs/Web/CSS/display#Values)
```
rtfeldman: Thanks! 👍

rtfeldman:
This is lovely! Thanks @folkertdev!

chaughawout:
I think that this merge closes Issue #183

Issue #204, opened 2016-12-30
Nested max/min-width/height media queries


jaredklewis:
Hello. The tutorial shows how to use nested media queries, such as `withMedia [ print ] [ color (hex "FF0000") ]`, but is it also possible to include the media's size? I would like to do something similar to the code below, which currently doesn't compile:

```
withMedia [ MediaQuery "max-width: 600px" ] [ color (hex "FF0000") ]
```

Of course if it is possible, it would also be great to have something like:

```
withMedia [ queryMaxWidth (px 600) ] [ color (hex "FF0000") ]
```

jaredklewis:
Ok, I've added support for these features (and many others!) in PR #207

jaredklewis:
After some further experimentation, I had another thought. One alternative to using media queries is using the `width` and `height` signals exposed by `Window` to conditionally apply inline styles.

This is not a total replacement of media queries as:

- Elm `Window` does not expose `window.screen` properties like `colorDepth` and `pixelDepth`.

- In order for media features like `grid` or `any-pointer` to work, Elm `Window` would need to expose the results of running `matchMedia` as signals to let Elm code be conditional upon those things.

- For external stylesheets, the typical CSS organization changes a bit. You need to define classes for each supported media arrangement (e.g. `ContainerXs`, `ContainerLg`, etc...) and then conditionally assign the class at runtime. In usual CSS, there is just a single `container` class that responds.

Thoughts anyone? Any Elm veterans out there that have tried sans-media query and used `Window.dimensions` only?

jaredklewis:
Another thought. Some of the more static media features such as `grid`, `any-hover`, `color` and so on could be collected by in JavaScript and then passed in as flags to `programWithFlags, though this might breakdown if the user moves the window to a different monitor or something.

globaljake:
created a simple PR that helps solve this that just exposes a mediaString helper PR #281

rtfeldman:
Closing now that we have a first-class media query API!

PR #207, opened 2017-01-04
Add full media query support

jaredklewis:
## Overview

This pull request supplements the the elm-css DSL with full support for media queries. 

It takes code like this:

```elm
import Css exposing (..)
import Css.Media as Media exposing (only, media, withMedia, mediaQuery, withMediaQuery, screen, print, or, not, monochrome, maxColorIndex, minResolution, dpi, anyHover)
import Css.Elements exposing (body, li, a)
import Css.Namespace exposing (namespace)

css =
    (stylesheet << namespace "demo")
        [ body [ color (hex "555555") ]
          -- top level media queries, using DSL
        , media [ print, or, monochrome, or, maxColorIndex (int 256) ]
            [ body [ color (hex "000000") ] ]
        , (.) LoResImage [ display block ]
        , (.) HiResImage [ display none ]
          -- supports all of CSS 3 media features
        , media [ only, screen, minResolution (dpi 300) ]
            [ (.) HiResImage [ display block ]
            , (.) LoResImage [ display none ]
            ]
        , a
            [ color (hex "0000FF")
              -- nested DSL media queries
            , withMedia [ print ] [ color (hex "000000"), textDecoration none ]
            , hover [ fontWeight bold ]
              -- supports CSS Level 4 media query features
            , withMedia [ anyHover none ] [ fontWeight bold ]
            ]
        , (.) Container
            [ maxWidth (px 800)
              -- if you want, use string, similar to the current API
            , withMediaQuery [ "screen and max-width: 600px" ] [ maxWidth (px 500) ]
            ]
        , (.) SquareScreenWarning [ display none ]
          -- string at the top level are also ok
        , mediaQuery [ "screen and (aspect-ratio: 4/3)" ] [ (.) SquareScreenWarning [ display block ] ]
        ]
```        

And turns it into CSS like this:

```css
body {
    color: #555555;
}

@media print,
(monochrome),
(max-color-index: 256) {
    body {
        color: #000000;
    }
}

.demoLoResImage {
    display: block;
}

.demoHiResImage {
    display: none;
}

@media only screen and (min-resolution: 300dpi) {
    .demoHiResImage {
        display: block;
    }
}

@media only screen and (min-resolution: 300dpi) {
    .demoLoResImage {
        display: none;
    }
}

a {
    color: #0000FF;
}

a:hover {
    font-weight: bold;
}

@media (any-hover: none) {
    a {
        font-weight: bold;
    }
}

@media print {
    a {
        color: #000000;
        text-decoration: none;
    }
}

.demoContainer {
    max-width: 800px;
}

@media screen and max-width: 600px {
    .demoContainer {
        max-width: 500px;
    }
}

.demoSquareScreenWarning {
    display: none;
}

@media screen and (aspect-ratio: 4/3) {
    .demoSquareScreenWarning {
        display: block;
    }
}
```

Apologies than I've suddenly made a rather big PR without consulting you before hand, @rtfeldman . After filing issue #204 , I started adding a little hack to my program to allow media queries and I was having so much fun playing around with elm that before I knew it, I had added a Media module, and thought I might as well throw in some tests and docs and PR this. 😄 

I would greatly appreciate any feedback or thoughts on how the PR could be improved. I'm open to all suggestions.

The PR includes tests and doc comments for all features.

## Background

I used primarily three references when writing this PR.

- [Media Queries Level 3](https://drafts.csswg.org/mediaqueries-3/)
- [Media Queries Level 4](https://drafts.csswg.org/mediaqueries/)
- Mozilla's concise [Using media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#aspect-ratio)

Basically CSS3 features are implemented in all browsers. CSS4 features are largely unimplemented, but some features have been implemented in some browsers (for example, the [interaction features](http://caniuse.com/#feat=css-media-interaction)). The Level 4 queries are not a finalized W3C standard yet.

This PR covers all CSS3 media query functionality, and includes support all CSS4 media features in the current working draft.

Level 4 includes an improved syntax for composing queries, that includes comparison operators (i.e. `width >= 300px`) and the ability to flexibly use `and`, `not`, and a new `or` keyword throughout the query.

The DSL in this PR however imitates the Level 3 syntax, not level 4.

## Composition

The API lets you build up media queries through a list.

    @media only screen and (min-width: 300px) {}

Becomes:
    
    media [ only, screen, minWidth (px 300) ] []

Although I find it is not often used, CSS3 lets you specify logical ORs by using a comma. Since we can't use a comma, I used `or`.

    @media (max-width: 300px), (max-height: 300px) {}

Becomes:

    media [ maxWidth (px 300), or, maxHeight (px 300) ] []

This PR's API allows construction of invalid media queries. The query below compiles.

    media [ maxWidth (px 300), only, only, screen, screen ]

This could be avoided by making an api with specified arguments like:

    media : MediaModifier -> MediaType -> List MediaFeature

But the problem there is that for the majority of queries, no modifier or media type is needed. I have never used an `only` modifier (which is only there to gracefully degrade legacy browsers.) The `not` modifier is useful for some situations, but usage is relatively rare.

With the exception of `print` and `speech`, media types serve little purpose as their qualities are better reflected in media features. Indeed, the Level 4 spec removes all media types except `all` (default, optional), `screen` (no point in use, well reflected in features), `speech`, and `print`. It also states of the goal of eventually removing all media types.

With that in mind, I went with a single List.

## Module Separation, `Mixin`s vs `MediaQueryComponent`s

There are some naming conflicts between CSS properties and media queries.

- The `color` media feature conflicts with the `color` property
- The `all` media type conflicts with the `all` property.
- The `minWidth`, `width`, `maxWidth`, `minHeight`, `height`, and `maxHeight` media features conflict with their corresponding properties.
- The `not` media modifier conflicts with the `not` pseudo class (although this is not implemented in elm-css yet). It also conflicts with `Basics.not`
- The `hover` media feature conflicts with the `hover` pseudo class

The property functions return `Mixin`s. Media features functions return `MediaQueryComponents`. This ensures correct usage properties and media features at compile time.

Maybe in the future, a more general type that encompasses both mixins and media features can be made, with later runtime analysis to see if it makes sense, with output of warnings of elm crashes to indicate poor usage. This would allow a unified DSL, though it would reduce compile time protections and would require more refactoring of existing code.

So, either the media feature functions need to be prefixed (like `mediaWidth`) and so on or separated into a different module.

With some hesitation I put them in separate module, though I am not against refactoring to use prefixes, and am open to other suggestions. 

Though it is a separate module, it still depends on many `Css` values like `none`, `scroll`, `px`, `pt`, `int` and so on

## Other Notes

- The `dpi`, `dpcm`, and `dppx` units can also be used in the `image-resolution` CSS property. This property is not implemented in browsers or elm-css yet, but is part of the CSS spec.

- I didn't understand some of the overloading magic. Perhaps someone could help me refactor the `hover` media feature/value pair to be overloaded? Currently I am using `hover canHover` as a hack, but the actual spec specifies this: `hover: hover`.

- Some media features can be either boolean or have a value. For example `@media (color) {...}` or `@media (color: 24) {...}`. Mostly I implemented these to accept an argument, except for `color` and `monochrome` where I let them be boolean features since in practice they almost always used in booleans. I suppose we could add `color1` and `monochrome1` but I can't imagine who would use them.

### DIFF
--- a/src/Css/Media.elm
+++ b/src/Css/Media.elm
@@ -1,0 +1,1168 @@
+module Css.Media
+    exposing
+        ( MediaQueryComponent
+        , MediaQuery
+        , media
+        , mediaQuery
+        , withMedia
+        , withMediaQuery
+        , only
+        , not
+        , or
+        , all
+        , screen
+        , print
+        , speech
+        , braille
+        , embossed
+        , handheld
+        , projection
+        , tty
+        , tv
+        , aural
+        , minWidth
+        , width
+        , maxWidth
+        , minHeight
+        , height
+        , maxHeight
+        , Landscape
+        , Portrait
+        , landscape
+        , portrait
+        , orientation
+        , Ratio
+        , ratio
+        , minAspectRatio
+        , aspectRatio
+        , maxAspectRatio
+        , Resolution
+        , dpi
+        , dpcm
+        , dppx
+        , minResolution
+        , resolution
+        , maxResolution
+        , Progressive
+        , Interlace
+        , progressive
+        , interlace
+        , scan
+        , grid
+        , Slow
+        , Fast
+        , slow
+        , fast
+        , update
+        , Paged
+        , OptionalPaged
+        , paged
+        , optionalPaged
+        , overflowBlock
+        , overflowInline
+        , Bits
+        , bits
+        , minColor
+        , color
+        , maxColor
+        , minColorIndex
+        , colorIndex
+        , maxColorIndex
+        , minMonochrome
+        , monochrome
+        , maxMonochrome
+        , SRGB
+        , P3
+        , Rec2020
+        , srgb
+        , p3
+        , rec2020
+        , colorGamut
+        , Fine
+        , Coarse
+        , fine
+        , coarse
+        , pointer
+        , anyPointer
+        , CanHover
+        , canHover
+        , hover
+        , anyHover
+        , InitialOnly
+        , Enabled
+        , initialOnly
+        , enabled
+        , scripting
+        )
+
+{-| Functions for composing
+[media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries).
+
+# Basics
+@docs MediaQueryComponent, MediaQuery
+
+# Query constructors
+@docs media, withMedia, mediaQuery, withMediaQuery
+
+# Query composition
+@docs only, not, or
+
+# Media Types
+@docs all, screen, print, speech, aural, braille, embossed, tv, tty,
+      handheld, projection
+
+# Viewport/Page Dimensions Media Features
+@docs minWidth, width, maxWidth, minHeight, height, maxHeight, Ratio, ratio,
+      minAspectRatio, aspectRatio, maxAspectRatio, Landscape, Portrait,
+      landscape, portrait, orientation
+
+# Display Quality Media Features
+@docs Resolution, dpi, dpcm, dppx, minResolution, resolution, maxResolution,
+      scan, Progressive, Interlace, progressive, interlace, scan, grid, Slow,
+      Fast, slow, fast, update, Paged, OptionalPaged, paged, optionalPaged,
+      overflowBlock, overflowInline
+
+# Color Media Features
+@docs Bits, bits, minColor, color, maxColor, minMonochrome, monochrome,
+      maxMonochrome, minColorIndex, colorIndex, maxColorIndex, SRGB, P3,
+      Rec2020, srgb, p3, rec2020, colorGamut
+
+# Interaction Media Features
+@docs Fine, Coarse, fine, coarse, pointer, anyPointer, CanHover, canHover,
+      hover, anyHover
+
+# Scripting Media Features
+@docs InitialOnly, Enabled, initialOnly, enabled, scripting
+
+-}
+
+import Css
+import Css.Preprocess as Preprocess exposing (Snippet, unwrapSnippet)
+import Css.Structure as Structure exposing (..)
+
+
+-- {- TODO: docs }
+{--Basics--}
+
+
+{-| The basic unit of a media query. Includes all media modifiers,
+media types, media features, and the `or` separator.
+Components are put together in lists, which are given to the media function.
+-}
+type alias MediaQueryComponent =
+    Structure.MediaQueryComponent
+
+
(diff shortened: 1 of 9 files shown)


tests/Media.elm
```diff
@@ -0,0 +1,405 @@
...
+                }
+            }
+            """
+    in
+        describe "withMediaQuery test"
+            [ test "pretty prints the expected output" <|
+                \_ ->
+                    outdented (prettyPrint input)
+                        |> Expect.equal (outdented output)
+            ]
```
rtfeldman: Great tests! 💪

src/Css/Media.elm
```diff
@@ -0,0 +1,1168 @@
...
+        , or
+        , all
+        , screen
+        , print
+        , speech
+        , braille
+        , embossed
+        , handheld
+        , projection
+        , tty
```
rtfeldman: `tty`, `tv`, `projection`, `handheld`, `braille`, `embossed`, and `aural` [were all deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/@media#Media_types) and "shouldn't be used" apparently. 😄

jaredklewis:
Hi @rtfeldman 

I really appreciate the feedback. I will take a look and push some changes soon.

eeue56:
@jaredklewis ping 😄

jaredklewis:
@eeue56  Sorry, still working on this. Got busy with work and such. 😄 

I will push my changes and rebase this week.

src/Css/Media.elm
jaredklewis: @rtfeldman Sorry for like the 3 month later followup here. Given their deprecation, do you think I should I remove them entirely or do you like you did for some other deprecated CSS things (but not all deprecated things) and have them be available, but unable to ever compile?

YetAnotherMinion: I vote for remove them. They are in Git history if anyone wants them later

splodingsocks:
@jaredklewis any progress on this?

src/Css/Media.elm
rtfeldman: yeah I'd say remove them entirely 👍

tolgap:
@rtfeldman @jaredklewis can we close this, as it was recently merged with #300 ?

rtfeldman:
@tolgap yep! Closing.

Issue #139, opened 2016-07-09
SVG elements


architectcodes:
Hey there, first of all warm thanks for creating this amazing tool!

Looking for a workaround for https://github.com/rtfeldman/elm-css-helpers/issues/2, I wanted to reach my `<svg>` through a `Css.children` selector combined with a `Css.Elements.svg`. But it turns out that `Css.Elements` only contains strict HTML elements.

I was wondering – is support for SVG elements on your roadmap?

rtfeldman:
I don't have any concrete plans for SVG elements, but offhand I can't think of why we couldn't support them!

Deimos28:
Hi! Thanks for the work on elm-css! I too would be very interested in styling SVGs!

rtfeldman:
Closing now that we have support for styling SVGs!

PR #216, opened 2017-01-18
Added SVG basic elements

wittjosiah:
Noticed #139 mentioned a desire for SVG elements. I've been using a couple of these on a project for a couple months, figured I'd fill out the rest of the basic shapes here. Not sure if there are other specific elements people might need, if there are just let me know and I can add them.

I also added `fill` property and `blockquote` element here.


rtfeldman:
Thanks @wittjosiah!

PR #225, opened 2017-02-06
Support nested modules

### LINKED ISSUE #218: Nested module names aren't supported
Currently one can customize the module name used by elm-css with the `--module=Whatever` option. The value passed in is used as a single string and a check is performed to look up the module name on the top-level Elm object:

```
var worker = Elm[stylesheetsModule].worker();
```

This means that any module name with nested qualifiers will fail. For example, trying to pass `Apps.Editor.Stylesheets` will result in

```
Error: I couldn't find the entry module Apps.Editor.Stylesheets.

Maybe you meant: Apps
You can pass me a different module to use with --module=<moduleName>
```

due to these lines:

```
if (!(stylesheetsModule in Elm)){
  return reject(missingEntryModuleMessage(stylesheetsModule, Elm));
}
```

A potential fix would have to walk the module name hierarchy both when looking up the module to create the worker and in the `suggestModulesNames` function for printing available modules when the module given truly isn't there

pauldijou:
Fix #218

### DIFF
--- a/index.js
+++ b/index.js
@@ -153,28 +153,50 @@ I was expecting a List CssFileStructure but got ${typeof portValue}!
   return errorMessage.trim();
 }
 
-function extractCssResults(dest, stylesheetsModule, stylesheetsPort) {
+function extractCssResults(dest, stylesheetsModuleName, stylesheetsPort) {
   return function () {
     return new Promise(function (resolve, reject) {
       var Elm = require(dest);
 
-      if (!(stylesheetsModule in Elm)){
-        return reject(missingEntryModuleMessage(stylesheetsModule, Elm));
+      /*
+      If you have a nested stylesheet Elm module like "My.Nested.Stylesheet"
+      the resulting Elm object will be like that:
+
+      Elm = {
+        My: {
+          Nested: {
+            Stylesheet: { worker: function () {...} }
+          }
+        }
       }
 
-      var worker = Elm[stylesheetsModule].worker();
+      So we need to split the module name on each dot and iterate
+      over the nested objects to reach the targeted one,
+      starting with the Elm object itself
+      */
+      var stylesheetsModule = Elm;
+      var parts = stylesheetsModuleName.split('.');
+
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -157,12 +157,17 @@ function extractCssResults(dest, stylesheetsModule, stylesheetsPort) {
...
-      }
+      parts.forEach(function (part) {
+        if (!worker || !worker[part]){
+          return reject(missingEntryModuleMessage(stylesheetsModule, Elm));
+        }
+        worker = worker[part];
+      })
 
-      var worker = Elm[stylesheetsModule].worker();
+      worker = worker.worker();
```
eeue56: This code is pretty confusing. Can I ask that you add comments, or at least simplify it?

pauldijou: Done

tolgap:
Rebasing with master should fix these build errors 👍

index.js
```diff
@@ -153,28 +153,39 @@ I was expecting a List CssFileStructure but got ${typeof portValue}!
...
 
-      if (!(stylesheetsModule in Elm)){
-        return reject(missingEntryModuleMessage(stylesheetsModule, Elm));
-      }
+      // Reach the targeted Elm module by iterating over the different sections
+      // of its name separated by dots
+      var parts = stylesheetsModuleName.split('.');
+
+      // The first part apply to the Elm object itself
+      var stylesheetsModule = Elm[parts.shift()];
```
eeue56: This comment doesn't really explain what the first part is meant to be

index.js
```diff
@@ -153,28 +153,39 @@ I was expecting a List CssFileStructure but got ${typeof portValue}!
...
+
+      // but the following ones will be nested objects inside the returned one
```
eeue56: an example will go a long way to making this easier to follow 👍

pauldijou: Done

rtfeldman:
@eeue56 any other thoughts here? I think this is 👍 to merge at this point.

PR #227, opened 2017-02-17
Proof of concept: @keyframes {...}

Warry:
Ongoing work, I'm not entirely sure I'm doing this right. See comments.


src/Css/Preprocess/Resolve.elm
```diff
@@ -141,11 +142,42 @@ resolveFontFace fontFaceProperties =
...
+            if step < 0 || step > 100 then
+                -- keyframe is dropped
+                extractWarningsFromKeyframes tl
+                    ( "Keyframe must be between 0 and 100 (those are percents)." :: warnings
+                    , steps
+                    )
+            else
+                let
+                    props =
+                        Preprocess.toPropertyPairs mixins
```
Warry: This is kind of brutal: converting a `Mixin` to properties without warnings.

tests/Animations.elm
```diff
@@ -0,0 +1,40 @@
...
+
+
+animations : Test
+animations =
+    describe "Animations"
+        [ test "keyframes" <|
+            \() ->
+                stylesheet
+                    [ keyframes "foo"
+                        [ ( 0, [ backgroundColor red ] )
```
Warry: alternative suggestion : 

```elm
keyframes "foo"
    [ start [ backgroundColor red ]
    , progress 33.3 [ backgroundColor blue ]
    , end [ backgroundColor yellow ]
    ]

```

Warry:
It's kind of working but now I'd like to understand what you mean about **warnings** and how you imagine I should deal with them.

The hart of the problem is here I think : https://github.com/nomalab/elm-css/blob/master/src/Css/Preprocess/Resolve.elm#L156-L181 where we "resolve" the keyframes.

I imagine that I should do a better job at converting `Preprocess.Mixin` to `Structure.Property` and detect misusages like nesting snippets etc... Do you agree ?

Warry:
Can someone please give me some feedbacks on this, so that I can complete the work?

kuon:
Any update on this? Full animation support in elm css would be very good. I can help too.

shuhei:
It seems like a related issue was created https://github.com/rtfeldman/elm-css/issues/313

PR #230, opened 2017-02-26
Add basic support for linear-gradient()

mthadley:
Hello everyone,

I was trying out `elm-css` on one of my projects and needed to use a `linear-gradient`. I wasn't able to find it anywhere so I thought I'd try adding it!

I attempted to follow the pre-existing patterns, however I think I may have missed or misinterpreted a few things. In particular, I'm not sure if `ColorStop`/`stop`/`stopAt` is exactly what we want.

Thanks!


.travis.yml
```diff
@@ -47,5 +47,5 @@ install:
   - cd ..
 
 script:
-  - elm-format --validate .
+  - elm-format --validate src tests examples readme-example
```
eeue56: I'd make this change as part of a seperate PR.

mthadley: Yea, that makes sense. I just wanted to move this PR into the green.

mthadley: Sent as #232

rtfeldman:
Thanks @mthadley! I'm gonna rename `stopAt` to `stop2` but otherwise, looks great! 🤘

PR #235, opened 2017-03-02
Calc

robrtle:
Adds basic support for calc()

```
   width (calc (pct 100) minus (px 2))

   width (calc (pct 100) minus (calc (pct 100) minus (px 2)))
```
Design choices 
--------------------

In my experience the dominant use-case for calc is two arguments (eg. compensating for a border). So I choose to represent the common case as concisely as possible. Calc can be nested for more then two arguments.

There should be no new opportunities to accidentally output invalid Css. 

I did not cover product expressions (  "*", "/"  ). They are redundant given elm-css's current functionality and add complexity for the more useful use-cases.

I also assume the plus and minus names are not going to be used elsewhere.


scottcorgan:
This is great!

Do you think it would be better to keep the functional order consinstent?

For example:

` width (calc <| minus (pct 100) (px 2))`

robrtle:
My thought was to "Make it look like the css"

'minus' and 'plus' could easily do the work as functions. It is more elm-like but might be harder for people to discover?

```
width (minus (pct 100) (px 2))

width (plus (pct 100) (px 2))

```

scottcorgan:
@robrtle I like it that way. Not sure what others will think.

robrtle:
Another option here is expanding the types for the |+| |-| |*| |/| operators.

This is however a breaking change.

```
{-| Add two lengths.

    >>> em 2 |+| em 3
    em 5
-}
(|+|) : Length compatible units -> Length compatible units -> Length compatible units

{-| Subtract two lengths.

    >>> em 7 |-| em 3
    em 4
-}
(|-|) : Length compatible units -> Length compatible units -> Length compatible units


{-| Divide two lengths.

    >>> em 9 |/| em 2
    em 4.5
-}
(|/|) : Length compatible units -> Length compatible units -> Length compatible units

{-| Multiply two lengths.

    >>> em 3 |*| em 6
    em 18
-}
(|*|) : Length compatible units -> Length compatible units -> Length compatible units
```

Would be more useful and consistent for Css as 

```
{-| Add two lengths.

    >>> em 2 |+| px 3
    calc (2em + 3px)
-}
(|+|) : Calc compatibleA -> Calc compatibleB -> CalculatedLength

{-| Subtract two lengths.

    >>> em 7 |-| em 3
    calc (7em - 3em)
-}
(|-|) : Calc compatibleA -> Calc compatibleB -> CalculatedLength


{-| Divide a length by a number.

    >>> em 9 |/|  2
    em 4.5
-}
(|/|) : Calc compatible -> number -> CalculatedLength

{-| Multiply a length by a number.

    >>> em 3 |*|  6
    em 18
-}
(|*|) : Calc compatible -> number-> CalculatedLength
```

rtfeldman:
Cool!

After taking a look at the APIs you've mentioned, I think the one in OP makes the most sense.

One of my design principles for this library is that CSS knowledge should translate to `elm-css` knowledge (and perhaps vice-versa). I like that for the typical use case, you can just figure out.

Here are some things I'd like to see:

* `times`
* `divideBy`
* Examples in the docs, especially for the "nested `calc`" case - what this does is reasonable, but is probably not what people would expect it to do. I would link to the MDN docs where it says that "nested `calc` is just treated as parentheses" so people understand that you're just doing this automatically for them.
* More tests - this covers a few cases, but I'd like to see more combinations of valid compositions with these, including various combinations of `plus`, `minus`, `times`, and `divideBy`

Thank you for this! 😄

robrtle:
A problem here is that divide and times require the second argument to be a number, not a length. 

so the options are

A function for each expression
```
width (minus (pct 100) (px 2))
width (plus (pct 100) (px 2))
width (divideBy (pct 100)  2)
width (times (pct 100)  2)
```

A 'calc' function for sums and a 'calc2' function for products
```
width (calc (pct 100) plux (px 2))
width (calc (pct 100) minus (px 2))
width (calcProduct (pct 100) divideBy 2)
width (calcProduct (pct 100) times  2)
```

An operator for each expression
```
width (pct 100 <+> px 2)
width (pct 100 <-> px 2)
width (pct 100 </> 2)
width (pct 100 <*>  2)
```
The operator approach seems the most reasonable to me if we need to include divideBy, times. By picking `<_>` instead of reusing `|_|` there is no backwards incompatibility issue. 

The current  `|*|` `|/|` could be removed eventually. 
`2px * 2px = 4(px^2)` does not really look right to me.

rtfeldman:
In CSS, is there a difference between `width: calc(100% / 2)` and `width: 50%`?

If not, it would suggest that the only reason `calc` supports these is for the benefit of handwritten `.css` files, which can't otherwise do arithmetic.

We don't have that problem, since we can do our arithmetic in Elm. If this is right, then `times` and `divideBy` don't actually do anything for us, and there's no reason to include them in the API. 😄

robrtle:
Yeah that's why I excluded them - I'll get the testing and documentation fixed :)

rtfeldman:
@robrtle Friendly ping re: tests and docs. I'd love to get this merged! 🤘

rtfeldman:
Awesome, thank you so much @robrtle!

PR #237, opened 2017-03-07
Add :visited pseudo-class

ahstro:
Add the :visited pseudo-class, useful for when you want to differentiate from links that have been visited and ones that have not.


rtfeldman:
Lovely, thank you @ahstro!

PR #238, opened 2017-03-07
Add brief reasons for different elm-css usage approaches

BobWhitelock:
I recently started using `elm-css`, and lacking a reason to choose one approach over the other I defaulted to using the first approach. After briefly discussing the differences with @rtfeldman on the Elm Slack a
while back, I decided that the second approach would better suit me. This is an attempt to briefly document the functional differences between the two approaches to hopefully help others make this choice.

I've also added some formatting while editing the `README`. Thanks!


rtfeldman:
Nice, thank you @BobWhitelock!

PR #239, opened 2017-03-08
Add ::-webkit-scrollbar pseudo-element

ahstro:
Add the `::-webkit-scrollbar` pseudo-element, which can be used to stylethe scrollbar in WebKit-based browsers.

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -295,6 +295,7 @@ module Css
         , firstLetter
         , firstLine
         , selection
+        , webkitScrollbar
         , active
         , any
         , checked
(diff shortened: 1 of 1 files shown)


rtfeldman:
Hi @ahstro,

Sorry, but I'm not okay with adding browser-specific selectors to the library. As the linked MDN article says:

> This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

ahstro:
Kind of figured, but worth a shot :)

PR #242, opened 2017-03-12
Add functions for custom pseudo-elements and pseudo-classes

ahstro:
First off, if this was already a thing and I just couldn't figure it out, my bad, just close this :)

This adds and exposes two functions - `pseudoElement` and `pseudoClass` - that teach take a `String` and a `List Mixin` and returns a `Mixin`. This lets people use custom elements and classes that aren't included in `elm-css` for some reason.

I also refactored the current definitions to use these functions instead.

Kind of a response to #239 being closed, but it also feels like something that should be available regardless. :)

Hope the tests were satisfactory (see "extant") this time? I figured that the current tests already tested the refactoring, so I just added tests for using the functions explicitly.

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -290,6 +290,7 @@ module Css
         , super
         , textTop
         , textBottom
+        , pseudoElement
         , after
         , before
         , firstLetter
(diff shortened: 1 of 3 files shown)


rtfeldman:
Lovely! I dig it.

Thanks @ahstro!

PR #249, opened 2017-03-27
Fix borderRadius4

ahstro:
The type signature was incorrect (didn't allow mixing length units on the first two arguments, so e.g., `borderRadius4 (em 0.2) (px 3) (px5) (px5)` was invalid.


rtfeldman:
Nice catch! Thank you @ahstro, and thanks for the test!

PR #250, opened 2017-03-27
Add borderWidth functions

ahstro:
Add borderWidth through borderWidth4 functions, for using the `border-width` property without specifying side.


rtfeldman:
Great, thanks @ahstro!

PR #252, opened 2017-04-03
Add inlineFlex for `display: inline-flex`

adamdicarlo:
This PR adds `inlineFlex` and runs `tests/Properties.elm` through the newest `elm-format`.

How's this look?


adamdicarlo:
Oh, I tracked down the build error: `elm-use` downloads `elm-format` from [bintray](https://dl.bintray.com/elmlang/elm-format/), which does not have the latest `elm-format` on it.

I'll pull out the commit that reformats.

rtfeldman:
Love it, thanks @adamdicarlo!

PR #253, opened 2017-04-13
Rename `Mixin` to `Style`

rtfeldman:
It's always been pretty weird that `backgroundColor` returns something called `Mixin`. My thinking was "these work like Mixins from Sass/Stylus/etc" but really what I want to say is just "these compose" and calling them Mixins seems unnecessary in retrospect.

What I should have said in the first place was "`backgroundColor` returns a `Style`, and you can compose `Style`s to make other `Style`s"

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -4,7 +4,7 @@ module Css
         , asPairs
         , Stylesheet
         , Snippet
-        , Mixin
+        , Style
         , Color
         , MediaQuery
         , Length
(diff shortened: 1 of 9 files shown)


mordrax:
w.r.t this PR, the following err now makes alot of sense:

```
The type annotation for `children` says it is a:

    List Preprocess.Snippet -> Style

But the definition (shown above) is a:

    List Preprocess.Snippet -> Css.Preprocess.Mixin
```

PR #254, opened 2017-04-15
Add Examples for Colors

k15a:
This will look like this:

bildschirmfoto 2017-04-15 um 11 25 38


rtfeldman:
Love it! 😻 

Thanks @k15a!

Issue #244, opened 2017-03-16
Missing "none" for "text-transform"


listrophy:
`none` is a supported value per the [spec](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform), but is not supported by elm-css.

## Expected Usage

    div [ textTransform none ]

rtfeldman:
Anyone want to take a stab at this? It's a pretty easy first contribution; the change should look very similar to https://github.com/rtfeldman/elm-css/pull/236

BobWhitelock:
I can take a look at this, I should be able to later today.

PR #256, opened 2017-04-17
Support more `text-transform` values

BobWhitelock:
This PR:

- adds tests for the `text-transform` properties already supported;

- adds support for the `none` property for `text-transform` (fixing https://github.com/rtfeldman/elm-css/issues/244);

- adds support for `inherit`, `initial`, and `unset` properties for `text-transform`, which I noticed https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform also says should be supported.

Hope this helps :slightly_smiling_face:.


rtfeldman:
Excellent, thanks @BobWhitelock!

PR #268, opened 2017-05-22
Add dl, dt and dd elements

skitoo:
`dl`, `dt` and `dd` elements are missing

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dt
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dd

### DIFF
--- a/src/Css/Elements.elm
+++ b/src/Css/Elements.elm
@@ -204,6 +204,8 @@ pre =
 
 
 {-| Selector for a dl element.
+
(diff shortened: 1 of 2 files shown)


skitoo:
@rtfeldman done ! :)

rtfeldman:
Nice - thanks @skitoo!

PR #272, opened 2017-06-08
Added missing `menu`, `aside`, `time`, `q` elements.

akoppela:


akoppela:
Done. Why all other elements does not have links to the relevant MDN docs?

PR #277, opened 2017-06-12
Implement `touch-action` and its values

franckverrot:
This commit adds a new method, `touchAction`, and all the values
supported by the `touch-action` property [1]:

  * `panX`
  * `panLeft`
  * `panRight`
  * `panY`
  * `panUp`
  * `panDown`
  * `pinchZoom`
  * `manipulation`

As explained in this post "More Responsive Tapping on iOS" [2] on
the WebKit blog, "WebKit on iOS has a 350 millisecond delay before
single taps activate links or buttons. [...] Since double tapping is
defined as two taps within a short time interval (350ms), WebKit must
wait for this time interval to pass before we’re sure that the user
intended to tap only once.".

Thanks to `touch-action`, and more specifically when it is set to
`manipulation`, single taps can be dispatched immediately.

[1] https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action?v=control#Values
[2] https://webkit.org/blog/5610/more-responsive-tapping-on-ios/

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -614,6 +614,15 @@ module Css
         , calc
         , plus
         , minus
+        , touchAction
+        , panX
+        , panLeft
+        , panRight
+        , panY
+        , panUp
+        , panDown
+        , pinchZoom
+        , manipulation
         )
 
 {-| Functions for building stylesheets.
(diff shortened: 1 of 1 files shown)


rtfeldman:
Cool, thanks @franckverrot!

PR #279, opened 2017-06-20
Forced cleanup of temporary files.

akoppela:
Temporary files were created but was not cleaned up after. So I forced cleanup.


rtfeldman:
Thanks @akoppela!

rtfeldman:
Published as 0.6.1

PR #281, opened 2017-06-25
Add and expose mediaString helper

globaljake:
This PR exposes a mediaString helper for custom media queries when using withMedia:

```
withMedia [ mediaString "screen and ( max-width: 600px )" ]
```

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -428,6 +428,7 @@ module Css
         , maxWidth
         , media
         , mediaQuery
+        , mediaString
         , medium
         , middle
         , minContent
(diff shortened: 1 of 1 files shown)


globaljake:
The second commit is only to pass your travis-ci build for elm-format@0.6.1-alpha. When only first commit is applied it is compliant with elm-format@exp.

globaljake:
simplified and rebased for version 9.1.0

rtfeldman:
There's a new type-safe media query API in  https://github.com/rtfeldman/elm-css/pull/308 which I think means we no longer need this! 😄

Issue #264, opened 2017-05-05
Can't create an !important rule in inline style


duncanmalashock:
```
view : Model -> Html Msg
view model =
    let
        bg =
            backgroundColor (rgb 129 20 100) |> important
    in
        div
            [ style <| asPairs [ bg ] ]
            [ text "hi!" ]
```
I expected this would produce:
```
<div style="background-color: rgb(129, 20, 100) !important;">hi!</div>
```

But it produces this in the view of my Elm app (Chrome 57.0.2987.133):
```
<div>hi!</div>
```

Changing `bg` to 
```
bg =
  backgroundColor (rgb 129 20 100)
```
produces:
```
<div style="background-color: rgb(129, 20, 100);">hi!</div>
```

The use case is that I'm trying to build a JS-based embed with inline styles that can't be overridden by styles on the page that embeds it. `!important` rules on an inline style attribute are highest-priority, so this feature would be very helpful!

BobWhitelock:
I investigated this, and the correct list of CSS keys to values does seem to be produced when `!important` is used with `asPairs` (see https://github.com/rtfeldman/elm-css/pull/284).

In an actual app though I can reproduce this: https://ellie-app.com/3D4YPznPsNBa1/1. However, the same issue occurs if elm-css is not used at all, for example: https://ellie-app.com/3D5jYcFc6f9a1/2.

It looks like this is actually an issue in `elm-lang/html`: https://github.com/elm-lang/html/issues/99.

rtfeldman:
Closing in favor of https://github.com/elm-lang/html/issues/99

PR #284, opened 2017-07-01
Add test case for issue #264

bobwhitelock:
This covers the issue described in https://github.com/rtfeldman/elm-css/issues/264, and demonstrates that using `!important` does give the expected result when used with `asPairs`.

I wrote this while investigating that issue, to ensure that the correct result was being produced at that level when `!important` is used. As can be seen this already passes, so feel free to close this if you don't think it adds any value. Thanks


rtfeldman:
I like it! Thanks @bobwhitelock

PR #291, opened 2017-07-10
9.1 Release

rtfeldman:
Added some missing doc comments, temporarily reverted `whiteSpace normal` support in order to allow for a `MINOR` release. (I'll revert it back for a future 10.0.0 release.)

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -709,6 +709,8 @@ module Css
 @docs listStyle, listStyle2, listStyle3
 @docs linearGradient, linearGradient2, stop, stop2, toBottom, toBottomLeft, toBottomRight, toLeft, toRight, toTop, toTopLeft, toTopRight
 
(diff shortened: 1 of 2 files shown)


bobwhitelock:
Not sure if you normally do this at this stage, but does the new release need adding to the [releases](https://github.com/rtfeldman/elm-css#releases) list as well?

rtfeldman:
@bobwhitelock released as 9.1!

ento:
I can't guess what happened, but the elm-package.json tagged as 9.1.0 says its version is 9.0.0 https://github.com/rtfeldman/elm-css/blob/9.1.0/elm-package.json#L2

eeue56:
@ento You can tag a commit without actually commiting the right elm-package.json change. You just need to have a tag of that version published, it doesn't actually look at the remote elm-package.json, only the local.

rtfeldman:
@ento fixed 😅 - also not sure why that happened.

PR #294, opened 2017-07-26
Correct text-justify to justify (fixes #293)

### LINKED ISSUE #293: textJustify compiles to "text-justify" instead of "justify"
Source code:
```
textJustify : TextAlign a b
textJustify =
    prop1 "text-justify"
```

Why is the property called `textJustify` and not just `justify`?
And the value that it compiles into, `text-justify` doesn't seem to be valid css.

Can the original piece of code become:
```
justify : TextAlign a b
justify =
    prop1 "justify"
```

bobwhitelock:
`text-justify` is an invalid `text-align` value, but `justify` is valid and we did not support it before.

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -373,6 +373,7 @@ module Css
         , int
         , invalid
         , italic
+        , justify
         , justifyAll
         , justifyContent
         , kannada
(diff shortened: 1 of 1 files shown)


rtfeldman:
Ha, thanks @bobwhitelock!

Issue #298, opened 2017-08-09
Namespace are not applied to ids (classes are fine)


qwfy:
Step to reproduce: Create two files with contents taken from README.md (pasted below), and call `elm-css --output=. Stylesheets.elm` would produce an index.css whose content is:

```
... 
#Page {
... 
}
... 
```
expect: (as stated in README.md)
```
...
#dreamwriterPage {
... 
}
...
```
Versions:
```
$ elm --version
0.18.0
$ elm-css --version
0.6.1
```

The two source files are: (copied from README.md)
```
-- File: MyCss.elm
module MyCss exposing (..)

import Css exposing (..)
import Css.Elements exposing (body, li)
import Css.Namespace exposing (namespace)


type CssClasses
    = NavBar


type CssIds
    = Page


css =
    (stylesheet << namespace "dreamwriter")
    [ body
        [ overflowX auto
        , minWidth (px 1280)
        ]
    , id Page
        [ backgroundColor (rgb 200 128 64)
        , color (hex "CCFFFF")
        , width (pct 100)
        , height (pct 100)
        , boxSizing borderBox
        , padding (px 8)
        , margin zero
        ]
    , class NavBar
        [ margin zero
        , padding zero
        , children
            [ li
                [ (display inlineBlock) |> important
                , color primaryAccentColor
                ]
            ]
        ]
    ]


primaryAccentColor =
    hex "ccffaa"



-- File: Stylesheets.elm
port module Stylesheets exposing (..)

import Css.File exposing (CssFileStructure, CssCompilerProgram)
import MyCss


port files : CssFileStructure -> Cmd msg


fileStructure : CssFileStructure
fileStructure =
    Css.File.toFileStructure
        [ ( "index.css", Css.File.compile [ MyCss.css ] ) ]


main : CssCompilerProgram
main =
    Css.File.compiler files fileStructure
```

bruce:
I've also encountered this issue with the same versions of `elm` and `elm-css`.

rtfeldman:
Closing since namespace has been deprecated!

PR #301, opened 2017-08-10
Namespace IDs

bruce:
This resolves rtfeldman/elm-css#298.

According to the README, IDs should be namespaced. This changes `applyNamespaceToRepeatable` so that `IdSelector` values are handled accordingly, similar to `ClassSelector` values, and updates the related tests.

### DIFF
--- a/src/Css/Namespace.elm
+++ b/src/Css/Namespace.elm
@@ -23,7 +23,7 @@ applyNamespaceToRepeatable name selector =
             ClassSelector (name ++ str)
 
         IdSelector str ->
-            IdSelector str
+            IdSelector (name ++ str)
 
         PseudoClassSelector str ->
             PseudoClassSelector str
--- a/tests/Compile.elm
+++ b/tests/Compile.elm
@@ -102,7 +102,7 @@ dreamwriter =
               display: none !important;
             }
 
-            #Page {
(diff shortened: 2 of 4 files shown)


eeue56:
This seems good to me

rtfeldman:
So IDs used to be namespaced, and then we decided that didn't make sense because IDs must be globally unique, and namespacing them can potentially (silently) break either styles or operations that use id selectors (e.g. `setFocus`), if the namespace changes and not all locations get updated.

So the actual bug is that the README didn't get changed along with the implementation. 😬

bruce:
Whoops, that's what I get for submitting a PR the first day I use a package. (Ah well, it took five minutes and it was educational!)

Thanks for the explanation; makes sense. I'm having a blast using elm-css, so thanks. 🙂

qwfy:
If so, should[ `Html.CssHelpers.withNamespace`](http://package.elm-lang.org/packages/rtfeldman/elm-css-helpers/2.1.0/Html-CssHelpers#withNamespace) still returns `id`?

tolgap:
@rtfeldman 

@qwfy raises a good question here. A function called `withNamespace` shouldn't be returning `id`, as it does not append the namespace. 🤔

PR #304, opened 2017-08-17
Added 'white' predefined color as also defined at http://clrs.cc/

amarqueslee:
It's just a small change. Less than I should do to say thank you for such a helpful project!


rtfeldman:
Thanks @amarqueslee - looks good!

PR #305, opened 2017-08-23
Revise media queries

rtfeldman:
This revises the media query API a bunch.

Some things it does:

## Always emit `only`

Reading up on [`only`](https://stackoverflow.com/questions/8549529/what-is-the-difference-between-screen-and-only-screen-in-media-queries/14168210#14168210), its basic purpose is to prevent older browsers which do not support media queries from breaking. Having older browsers not break seems always desirable, so the only reason not to use `only` across the board is to save 5 characters of emitted CSS. Especially given how well `@media only` will gzip, I don't think that's enough to justify adding complexity to an already complex API.

## Remove deprecated media types

e.g. `tv`, `aural`, `embossed`...

## Rule out invalid combinations of `and` and `or`

Previously it was possible to do things like `[ or, or, or, or ]` - now `or` & `and` take exactly 2 media queries as arguments (and `not` takes 1 argument), so they can only be used in conjunction with existing queries.

## Parenthetical groupings for `and` and `or`

Now `and` and `or` emit parens around their arguments, so the output will have whatever parenthetical groupings your code indicates should be there.

### DIFF
--- a/src/Css/Media.elm
+++ b/src/Css/Media.elm
@@ -10,6 +10,7 @@ module Css.Media
         , Interlace
         , Landscape
         , MediaQuery
+        , MediaType
         , OptionalPaged
         , P3
         , Paged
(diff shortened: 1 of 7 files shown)


eeue56:
lgtm 👍

rtfeldman:
@eeue56 published as 10.0.0!

Issue #307, opened 2017-08-27
Media query API


mattjbray:
The type-safe media queries are a cool idea :)

However I think the generated CSS is incorrect.

For reference, here's the BNF (from [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#Pseudo-BNF), also a less readable version in the [CSS3 spec](https://drafts.csswg.org/mediaqueries-3/#syntax)):

```
media_query: [[only | not]? <media_type> [ and <expression> ]*]
           | <expression> [ and <expression> ]*
expression: ( <media_feature> [: <value>]? )
```

(I'm not an expert on media queries -- I've only read the linked MDN page -- so if I have this all wrong please ignore.)

## Case 1: Single expression

This snippet:

```elm
css =
    Css.stylesheet
        [ Media.media (Media.minWidth (px 768))
            [ body [ color (hex "f00") ] ]
        ]
```

generates:

```css
@media only (min-width: 768px) {
    body {
        color: #f00;
    }
}
```

which is incorrect because `only` may only appear before a `<media_type>`.

## Case 2: Specifying a media type

This snippet:

```elm
css =
    Css.stylesheet
        [ Media.media (Media.and Media.screen (Media.minWidth (px 768)))
            [ body [ color (hex "f00") ] ]
        ]
```

generates:

```css
@media only (screen and (min-width: 768px)) {
    body {
        color: #f00;
    }
}
```

which is incorrect because `(screen and ...)` is parenthesised - only individual expressions should be in parentheses.

## Case 3: Invalid specifications

It's also possible to specify invalid media queries in elm-css.

For example:

* Specifying two media types with `and` is not valid:

   ```elm
   Media.media (Media.and Media.screen Media.print)
   ```

   generates
   
   ```css
   @media only (screen and print)
   ```

* `Media.or` assumes an `or` keyword that doesn't exist (queries should be comma-separated instead):

   ```elm
   Media.media (Media.or Media.screen Media.print)
   ```
   
   generates
   
   ```css
   @media only (screen or print)
   ```

* Only top-level queries may be negated, but elm-css allows negating individual or nested expressions:

   ```elm
   media (and screen (not (minWidth (px 768))))
   ```

   generates

   ```css
   @media only (screen and not (min-width: 768px))
   ```

## Proposal

In the spirit of making impossible states impossible, the last case motivates a new API for media queries. I think this representation also makes the CSS code generation rules more straightforward.

```elm
[...]

PR #308, opened 2017-08-28
Media query API

mattjbray:
This implements #307.

There were no real surprises in the implementation.

I considered renaming `Expression` to `Feature` or `MediaFeature` -- technically `Expression` is the correct term, but `Feature` might be more intuitive. What do you think?

There is a choice in how we render `media [ all [color] ] [ ... ]`:

1. `@media (color) { ... }`
2. `@media only all and (color) { ... }`

Having read [this SO answer](https://stackoverflow.com/questions/8549529/what-is-the-difference-between-screen-and-only-screen-in-media-queries/14168210#14168210), I think both should be ignored by old browsers so it shouldn't matter. I went with option (1).

### DIFF
--- a/src/Css/Media.elm
+++ b/src/Css/Media.elm
@@ -88,9 +88,7 @@ module Css.Media
         , withMediaQuery
         )
 
-{-| Functions for composing.
-
-@docs Bits
+{-| Functions for building `@media` queries.
 
 
 # Data Structures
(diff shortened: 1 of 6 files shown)


rtfeldman:
Fantastic! 😻

Thank you so much @mattjbray!

rtfeldman:
Published as 11.0.0!

PR #309, opened 2017-08-31
Use Flow & prettier

rtfeldman:


dustinfarris:
Flow's license is BSD+Patents.  Does that now extend to elm-css, and, by further extension, to my project?  Just curious.

rtfeldman:
I found [a discussion](https://github.com/vuejs/vue/issues/6411#issuecomment-324701129) on this, and the conclusion is "that wouldn't make sense." That said, "wouldn't make sense" from anyone but a lawyer isn't worth much in my book.

This comment is making me wonder if I should just move it to TypeScript so I don't have to worry about this ever again. 😛

rtfeldman:
Although since TypeScript actually emits the code that gets published, it's actually more likely that TypeScript would carry an Apache2 License along with it. (That's far less burdensome than BSD+FBPatents, of course, but my preference is that `elm-css`'s BSD-3-Clause license is all there is to it.)

rtfeldman:
Thinking about it more, it would be hard to believe this would be a problem. After all, `gcc` is licensed under the GPL; it type-checks *and* emits code, and people have been using it to compile and emit non-GPL code for decades.

I can't find anything in Flow's `LICENSE` or `PATENTS` file that would lead me to believe differently, so my default thinking is that it's fine until I hear otherwise from a lawyer. 😄

PR #319, opened 2017-09-16
Fix generatedElmStuff check

ahstro:
Only check if not symbolic link if path exits


rtfeldman:
I believe this is correct as-is. I was seeing different behavior per-OS, where (as I recall, on Linux) Node reports "this file does not exist" if it's a symlink, so you have to check both that *and* then separately check that it's not a symlink.

Is there a bug you're seeing that this change would fix?

ahstro:
The problem I ran into was that if `elm-stuff/generated-code/rtfeldman/elm-css/elm-stuff` does not exist at all, `!fs.existsSync(generatedElmStuff)` will return `true`, causing `fs.lstatSync(generatedElmStuff)` to run on a non-existing path, which causes it to throw the exception below. This happens both on Linux and macOS for me. The solution to me seems to be to to only run `fs.lstatSync` like in my PR, where we know the path exists for sure, but it's very possible that there are edge cases I'm not thinking of 😅 

```
elm-css 0.7.0-beta
------------------

fs.js:948
  binding.lstat(pathModule._makeLong(path));
          ^

Error: ENOENT: no such file or directory, lstat '/home/ahstro/home/dev/ahstro/elm-stuff/generated-code/rtfeldman/elm-css/elm-stuff'
    at Object.fs.lstatSync (fs.js:948:11)
    at Object.lstatSync (/home/ahstro/home/dev/elm-css/node_modules/graceful-fs/polyfills.js:297:22)
    at module.exports (/home/ahstro/home/dev/elm-css/index.js:55:9)
    at Object.<anonymous> (/home/ahstro/home/dev/elm-css/js/cli.js:27:1)
    at Module._compile (module.js:624:30)
    at Object.Module._extensions..js (module.js:635:10)
    at Module.load (module.js:545:32)
    at tryModuleLoad (module.js:508:12)
    at Function.Module._load (module.js:500:3)
    at Function.Module.runMain (module.js:665:10)
```

rtfeldman:
This is now fixed in `master` with a try/catch so it works on both macOS and Linux!

PR #321, opened 2017-09-17
Access (what I hope is) the correct Css$class in hack-main

ahstro:
Not sure if this is correct, nor if it should be hardcoded, but this fixed this error for me (and let me successfully generate CSS, so I think we're good*ish*): `ReferenceError: _ahstro$ahstro$Css$class is not defined`


rtfeldman:
I think this should actually break things haha...`Css.class` comes from this package, so it should be namespaced the way it currently is.

Are you by chance including this library as a `source-directory` instead of as a `dependency`? :thinking:

ahstro:
That sounds very possible 😅

What do you mean by "comes from this package"?

I think I'm including things as they should be; `elm-stuff/generated-code/rtfeldman/elm-css/classes` as a `source-directory` in [my main `elm-package.json`](https://github.com/ahstro/ahstro.github.io/blob/1862bdccc11a03d476234b558bd208d053c56b6b/elm-package.json) and `rtfeldman/elm-css` as a `dependency` in [my `css/elm-package.json`](https://github.com/ahstro/ahstro.github.io/blob/1862bdccc11a03d476234b558bd208d053c56b6b/css/elm-package.json). Feel free to clone and build my project to see if it works for you :)

PR #322, opened 2017-09-18
Fix perspective transform.

mrnfor:
Perspective value should be of type length or none, as noted [here](https://developer.mozilla.org/en-US/docs/Web/CSS/perspective).


rtfeldman:
Nice, thanks @mrnfor!

PR #323, opened 2017-09-21
Update README.md

dustinfarris:
Replace mixin with style


rtfeldman:
Thanks @dustinfarris!

PR #326, opened 2017-10-08
Added property visibility

Pautrik:


rtfeldman:
Oh wow, I can't believe we went this long without having this one! :sweat_smile: 

Thanks @Pautrik!

Issue #295, opened 2017-08-03
Each does not keep pseudoclasses


Stijnkool:
The following code:
```elm
each
    [ class Test
    , \s -> class Test [ focus s ]
    , \s -> class Test [ focus [ hover s ] ]
    ]
    [ color (hex "FF0000")
    ]
```
compiles into:
```css
.Test,
.Test,
.Test {
   color: "#FF0000";
}
```
while I think it should compile into:
```css
.Test,
.Test:focus,
.Test:focus:hover {
   color: "#FF0000";
}
```
Or am I missing something?

elm-css version: `"9.1.0 <= v < 10.0.0"`

tolgap:
This is indeed a bug.

The culprit is `collectSelectors`: https://github.com/rtfeldman/elm-css/blob/d50ef4bc559d021c43ea30e046d460f45a6301f9/src/Css.elm#L8358

As you can see, the child styles are ignored with an `_`. We somehow would need to resolve those styles into `firstSelector`. @rtfeldman Is there an existing way to "merge" styles into a selector?

So something like `Structure.Selector -> List Style -> List Stucture.Selector`? I can't seem to find it, but it's something that would be needed to do this.

PR #329, opened 2017-10-18
Bugfix: Pseudoclasses in each

tolgap:
This should resolve issue #295 . This was quite a hard one to figure out, as I needed to read through almost all logic inside `Structure.elm` to figure out what was going on 😅 .

I haven't quite been able to figure out the crazy possibilities, like this nesting:

```
each
    [ span [ descendants [ span [ hover [ children [ color (hex "#000") ] ] ] ]
    ]
```

Nesting combinators inside selectors seems to not work yet (they get ignored now). That's because we would need to update a different "field" of a selector which we don't have access to when we're dealing with sequences.

PS:
Sequence = span:hover
Combinator = span span

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -8365,16 +8365,30 @@ collectSelectors declarations =
 unwrapSelector : Selector -> List Style -> Selector
 unwrapSelector (Structure.Selector sequence combinators mPseudo) styles =
     let
-        unwrapNestedSelector style s =
+        unwrapSequenceSelector style s =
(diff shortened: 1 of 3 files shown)


rtfeldman:
Thank you so much @tolgap! Amazing work as always. 😻 

I rebased this on `master` and merged it in manually: https://github.com/rtfeldman/elm-css/commit/458ea741af55e417dfec0ed88796e5e1f7622d0f

PR #331, opened 2017-10-20
Use foldr instead of foldl and reverse

ahstro:
Instead of using `List.foldl` to reduce children and then having to `List.reverse` the child nodes, use `List.foldr` and skip reversing.

I *think* this makes sense, since `Dict`s (the other thing being accumulated with the `fold*`) are often stated as a valid to use when order is irrelevant. I suppose there *might* be some potential problem if the order of the `Dict` is actually relied upon to order the styles properly (them being cascading and all). The example page looks the same after this change, but it's not very complex, so the likelihood of styles conflicting based on order is minimal.

### DIFF
--- a/src/Html/Styled/Internal.elm
+++ b/src/Html/Styled/Internal.elm
@@ -63,7 +63,7 @@ unstyle elemType attributes children =
             List.foldl accumulateStyles Dict.empty attributes
 
         ( childNodes, styles ) =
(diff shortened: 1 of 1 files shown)


rtfeldman:
@ahstro I'm concerned that this would actually cause a regression instead of working as intended 😅 - do you think you could add a test case demonstrating that it still works properly?

Issue #339, opened 2017-11-24
Svg.Styled


edkv:
`Html.Styled` is awesome, but are there any plans for adding `Svg.Styled`? I tried to implement it on top of `Html.Styled`, but this leads to runtime errors since `elm-css` adds classes using the [`className` property](https://github.com/rtfeldman/elm-css/blob/master/src/Html/Styled/Internal.elm#L315) and for Svg [the `class` attribute should be used instead](https://github.com/elm-lang/svg/blob/2.0.0/src/Svg/Attributes.elm#L228).

rtfeldman:
Ah, good point! You can I'm totally open to adding `Svg.Styled` modules.

Want to take a crack at a PR?

edkv:
@rtfeldman I'm implementing `Svg.Styled` and I've noticed that the `styled` function is defined in the `Html.Styled.Attributes` module. Is this really intended? It seems to me that `Html.Styled` is a better place for it since it works with nodes and not with attributes. Maybe we should move it?

rtfeldman:
👍 I agree! Go ahead and move it.

The fix for https://github.com/rtfeldman/elm-css/issues/337 (namely making it so that if you provide multiple `css` attributes to the same element, only one has any effect) should be a breaking release anyway.

rtfeldman:
Fixed by #342

PR #340, opened 2017-11-25
Replace Html.Styled.Internal with VirtualDom.Styled

edkv:
So, to introduce `Svg.Styled` (#339) we need at first to move `Html.Styled.Internal` functionality to some common place. This PR replaces it with the `VirtualDom.Styled` module which should now serve as a base for `Html.Styled` and `Svg.Styled`.

If the idea is OK, we should also discuss whether this module should be exposed. If yes, we should add documentation and clean up its API:  `makeSnippet` function should be moved to another place; `getClassname` function and `Property` constructor shouldn't be exposed. They are currently exposed because `Html.Styled.css` function needs to render classes as the `className` _property_ and `Svg.Styled.css` will need to use the `class` _attribute_, so we'll need to find a better way how to achieve that.

### DIFF
--- a/src/Css/Foreign.elm
+++ b/src/Css/Foreign.elm
@@ -164,8 +164,8 @@ import Css.Preprocess as Preprocess
 import Css.Preprocess.Resolve as Resolve
 import Css.Structure as Structure
 import Html.Styled
-import Html.Styled.Internal exposing (makeSnippet)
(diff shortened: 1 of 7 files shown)


rtfeldman:
Looks good, thanks @edkv!

We definitely should not expose this module. :smile:

Issue #337, opened 2017-11-22
Incorrect CSS cascade when styles compiled to attributes are mixed


Stijnkool:
When having a styled element, lets say a button.
```elm
styledButton : List (Attribute msg) -> List (Html msg) -> Html msg     
styledButton =
    styled button
        [ padding (px 10)
        , border3 (px 1) solid (hex "FF0000")
        ]
```
We can use it as it is (the base version):
```elm
view : Html msg
view =
    styledButton [] [ text "a button"]
```
But what if we want to extend the styled element with some other styles.
I want to keep the original styles but increase the padding. (the extending version)
```elm
anotherView: Html msg
anotherView =
    styledButton [ css [padding (px 20)] ] [ text "a bigger button"]
```
So I am adding 2 padding rules. The last one in the outputted css will be used. But different classes are being generated for each (for the base styles and for the extending styles).

- How can elm-css guarantee that the extending class will be after the base class (the class generated by the base css on the styled element) in the final css. 
- Should elm-css guarantee it? The order of the attributes should have no meaning.
- I am running into cases where the extending class is after the base class and vice versa
- This seems like a problem inherent with mixing and matching styles compiled to attributes.
- Is this the problem to what the line _This is less flexible than mixing and matching `Style` values_ refers to?

Thanks a lot. I love the `Html.Styled` way of doing things!

rtfeldman:
Thanks for bringing this up! Quick version of how this currently works:

* It doesn't matter how you order the attributes. You'll get the same output.
* If you use `css` attributes twice on the same element, and they both define the same property, there is no guarantee which definition will "win."

I hadn't realized this in the original design, but it's potentially easy to end up with surprising bugs from this. Like if I write the code you wrote above, and then later, I add a new property to one of the two styles, that might cause the order in which they get emitted to be flipped - which could effectively impact properties other than the one I added.

I'd be pretty surprised if that happened...I bet I wouldn't have thought to re-test those styles.

That said, it's possible to change the compiled output to make this no longer true. We could instead make it so that each subsequent attribute overrode the previous one - such that your example above would work.

To make this work, we'd have to compile to one class per declaration, the way [Styletron](https://github.com/rtsao/styletron) does. I have not looked into what the performance implications of this change would be, but Styletron [claims to be faster than the hashed-class approach](https://ryantsao.com/blog/virtual-css-with-styletron), so maybe it's a win-win.

aleksdao:
[...]

PR #342, opened 2017-11-27
Move Html.Attributes.styled to Html.Styled module; add Svg.Styled modules.

### LINKED ISSUE #339: Svg.Styled

edkv:
Closes #339.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ module MyCss exposing (main)
 import Css exposing (..)
 import Html
 import Html.Styled exposing (..)
-import Html.Styled.Attributes exposing (css, href, src, styled)
+import Html.Styled.Attributes exposing (css, href, src)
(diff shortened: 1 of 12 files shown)


rtfeldman:
Looks great! 😍 

Thank you @edkv! I need to make one more change to fix https://github.com/rtfeldman/elm-css/issues/337 and then I can release this.

Issue #348, opened 2017-11-30
Html.Styled.Lazy API defeats VirtualDom.lazy's checks


edkv:
Here is the definition of the `lazy` function:
```elm
lazy : (a -> Html msg) -> a -> Html msg
lazy fn arg =
    VirtualDom.lazy (Html.toUnstyled << fn) arg
        |> Unstyled
```
The important part is that we create the function which we then pass to `lazy` on the fly with 
`Html.toUnstyled >> fn`.

I've just discovered that `VirtualDom.lazy` doesn't do any optimizations with functions that are not defined statically. See the following example: https://ellie-app.com/9r58v3htva1/0.

Only `model.value1` is passed to `lazy`.
```elm
        , lazy
            (\_ -> text (toString model))
            model.value1
```
But when we increment `model.value2`, the view is still updated.

[As I can see from the `VirtualDom`'s source](https://github.com/elm-lang/virtual-dom/blob/master/src/Native/VirtualDom.js#L576), it uses `===` to compare functions by reference. Since `Html.Styled.Lazy.lazy` creates a new function every time, [`if (same)`](https://github.com/elm-lang/virtual-dom/blob/master/src/Native/VirtualDom.js#L581) results to `false` and the view is always rebuilt.

If this is correct, `Html.Styled.Lazy` is useless with the current implementation. And I'm not sure if it can be implemented the other way...

rtfeldman:
Bah, good catch. The API needs to be:

```elm
lazy : (a -> VirtualDom.Node msg) -> a -> Html msg
```

...and then the caller has to call `toUnstyled` in the named function.

I just verified that this fixes it. Will make a PR.

Thanks for reporting this!

edkv:
Oh, I'm glad there is a workaround! Sad that it's no longer a drop-in replacement though.

PR #351, opened 2017-12-01
Fix lazy2 and lazy3

edkv:
@rtfeldman `lazy2` and `lazy3` probably still suffer from #348 since you've used anonymous functions there.


rtfeldman:
Good point 😅 - thanks @edkv!

PR #355, opened 2017-12-06
Fix divide by zero error in arithmetic tests

tolgap:
Every now and then, when I ran the arithmetic tests with a certain seed, it would fail because the 3rd number would be `0`.

There was a catch for the second number being `0`, but I also added it for the third number.

The test would do `number1 |/| number2 |/| number3 |/|`.


rtfeldman:
Lovely! Thanks @tolgap!

Issue #335, opened 2017-11-22
`linearGradient2` with degree direction value compiles to invalid CSS


maxhallinan:
`linearGradient2` takes a `AngleOrDirection` as its first argument. It then [prepends this value](https://github.com/rtfeldman/elm-css/blob/master/src/Css.elm#L4339) with `"to"`. When that value is a degree instead of a direction, the result is invalid CSS.

Example:

```
linearGradient2  (deg 10) (stop <| hex "000") (stop <| hex "222") []
```

compiles to

```
linear-gradient(to 10deg, rgba(209, 237, 188, 0.5), rgba(209, 237, 188, 0.25))
```

Am I missing something? This library would benefit from an overview of the domain model. Reverse engineering the domain model from type annotations is not efficient for new users. Because I'm never sure that I understand the domain model, I'm not sure that I'm approaching this problem the correct way. If this is a legitimate bug, I'm happy to work on a fix. Wanted to make sure first.

rtfeldman:
@maxhallinan Yeah, looks like a bug. I'd definitely appreciate a PR for a fix!

maxhallinan:
Thanks for taking a look. I'll work on a fix.

PR #356, opened 2017-12-08
Compile `AngleOrDirection` to correct CSS string

maxhallinan:
This fixes the bug described in issue #335  . The fix seems a little naive to me. It looks like a value of type `AngleOrDirection` will have an `angle` field when it is of type `Angle`. The other idea was to use the presence of this field on the `dir` argument to infer an angle or direction value . But since it looks like the built-in direction values are only used in the context of a gradient, adding `"to"` to the value was easier. Let me know if this assumption is incorrect. 

The `mocha *.js` part of `npm test` fails for me. The error message is `Elm compiler "/Users/maxhallinan/Stuff/elm-css/tests" did not have permission to run. Do you need to give it executable permissions?`. This test fails on master too. Unfortunately, I do not have time to debug this now but it does not seem related to the changes on this branch.

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -4336,7 +4336,7 @@ linearGradient2 dir stop1 stop2 stops =
         [ stop1, stop2 ]
             ++ stops
             |> collectStops
-            |> (::) ("to " ++ dir.value)
+            |> (::) dir.value
             |> cssFunction "linear-gradient"
     , backgroundImage = Compatible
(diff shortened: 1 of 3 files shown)


danilol:
I'm currently facing the same problem.

rtfeldman:
Great, thanks @maxhallinan!

rtfeldman:
Published as 13.1.1!

PR #357, opened 2017-12-14
Replace emitter test with elm-html-test by running it on readme-example with Html.Styled

tolgap:
This PR should fix all tests again.

The old emitter test is removed, because compiling to CSS files is deprecated. It also didn't even work (causing all PR builds to fail).

I simply source the `readme-example` and check each css class output. @rtfeldman is it bad practice to actually use the hashed css class in my tests? I think it's nice, because the test will always need to be determinate this way.


tolgap:
@rtfeldman Are Travis builds disabled?

rtfeldman:
> is it bad practice to actually use the hashed css class in my tests? I think it's nice, because the test will always need to be determinate this way.

I think it's fine - we'll have to manually update them if the algorithm changes, but that shouldn't happen often. 😄

rtfeldman:
Thanks @tolgap!

PR #376, opened 2017-12-28
linearGradient doc fix

2mol:
fixed mistyped function name for linearGradient2, removed % symbol from pct


rtfeldman:
Thanks @2mol!

Issue #247, opened 2017-03-26
pointer-events


mordrax:
I'd like "pointer-events" to be supported. I'm happy to [attempt a PR for this](https://github.com/rtfeldman/elm-css/issues/14), but since I've only started using elm-css moments ago, it would be good to be pointed to some similar style implementations so I can see how this works.

rtfeldman:
Thanks for the issue!

Unfortunately `pointer-events` is [still in the Recommendation stage](https://www.w3.org/TR/pointerevents/) and has been since 2015. Firefox and Safari do not support it at all.

I have to draw the line somewhere for what is in-scope to be covered, and this proposal is still too preliminary for me to justify including it. 😅

Tokap:
@rtfeldman 

Hi Richard, thanks for taking the time to review this issue. I definitely understand the need to draw a line on css properties not supported by the average browser. However, [Can I Use](https://caniuse.com) seems to suggest that the JS `window.pointerEvent` remains a suggestion, while the CSS `pointer-events` property is currently supported by all major browsers.

#### Javascript `window.pointerEvent` Information:
* [Can I Use](https://caniuse.com/#feat=pointer)
* [MDN](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent)
* [W3C](https://www.w3.org/TR/2015/REC-pointerevents-20150224/#examples)

#### CSS `pointer-events` Information:
* [Can I Use](https://caniuse.com/#feat=pointer-events)
* [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events?v=example)

Based on the above information:
1. The CSS property `pointer-events` is currently supported by all major browsers.
2. The JS `window.pointerEvents` is not supported by Firefox or Safari.
3. There seems to be a well known ambiguity between the JS and CSS implementation.  The [Can I Use](https://caniuse.com/#search=pointer) article alludes to the confusion caused by this ambiguity:
>"This specification integrates various inputs from mice, touchscreens, and pens ... . Not to be mistaken with the unrelated `pointer-events` CSS property."

Therefore, given the well-known confusion, I want to ensure we are all referring to the same topic.  @mordrax, @rtfeldman were you both referring to the CSS `pointer-events` property?  If not, would you please take another look at this issue?  

Thank you for all the great work and for your attention to this issue.

mordrax:
[...]

Issue #375, opened 2017-12-25
Phantom Values (contributors, please take note!)


rtfeldman:
I'm working on having `Css.Value` use phantom types. (More on what this means below.)

If you're using `elm-css`, you may not even notice this change when it comes out (except that you may notice things running faster), but if you're contributing to current `master`, this will almost certainly lead to merge conflicts - so watch out! Any large PR will likely need numerous changes once this is finished, and I would strongly recommend opening an issue before making a PR so we can coordinate.

You can follow progress on the [`phantom-values` branch](https://github.com/rtfeldman/elm-css/tree/phantom-values). This one is gonna take awhile to complete. 😅 

## The Change

On current `master`, `Css.Value` is defined like this:

```elm
type alias Value compatible =
    { compatible | value : String }
```

On the [`phantom-values` branch](https://github.com/rtfeldman/elm-css/tree/phantom-values) it's instead defined like this:

```elm
type Value a
    = Value String
```

A union type with a type variable which doesn't appear in any constructors is known as a [phantom type](https://medium.com/@ckoster22/advanced-types-in-elm-phantom-types-808044c5946d). I'm somewhat annoyed that they have such a cool name, because it makes me feel like I should use them more often...when in reality, they are useful under very rare circumstances.

However, `elm-css` happens to be one of them!

What the phantom type lets us do is move our compatibility information from runtime values to compile-time values. For example:

```elm
type alias ColorValue compatible =
    { compatible | value : String, color : Compatible }
```

This can now become:

```elm
type alias ColorValue compatible =
    Value { compatible | value : String, color : Compatible }
```

The difference is subtle, but impactful. We still get all the same compile-time verification as before, but whereas before we actually had to instantiate all those fields in real objects, the runtime representation of `Value` is now always a single `String` constructor: `type Value a = Value String`. All the compatibility checking information now exists at compile time only.

If that were the only part of the refactor, though, that wouldn't be a terribly big change. The big change is shifting around how the open records work.

## Much Nicer Error Messages

Credit to @ianmackenzie for showing me this. Switching around which records are open and which ones are closed can result in much nicer documentation and compile-time error messages for `elm-css`!

For example, let's consider how `color` and `rgb` interact.

#### Status Quo

Here's how these are defined right now.

```elm
color : ColorValue compatible -> Style

rgb : Int -> Int -> Int -> Color

type alias Color =
    ColorValue { red : Int, green : Int, blue : Int, alpha : Float }

[...]

PR #377, opened 2018-01-08
Add `pointer-events` property

gurdiga:
## Summary

Another try to solve https://github.com/rtfeldman/elm-css/issues/247. (First was https://github.com/rtfeldman/elm-css/pull/276 which seems to be abandoned because the conversation stopped on Jun 11, 2017. 🤔)

This is my first totally-shot-in-the-dark PR, and I’m open to guidance. 🙃 Is this close to how the **Right Way®**? 🤓

 ### The `format` NPM script

The initial build for this PR failed because I had a different version of `elm-format` installed globally.  😑

To prevent this for other contributors, I’ve also added the `format` NPM script — now at least there is some hint about the formatting requirements. And, for that script to do the right thing I’ve also added `elm-format@0.7.0-exp` to the `devDependencies` so that we don’t depend on the individual contributor’s `elm-format` version. 😎

NOTE: This is totally movable to a separate PR if it’ll fit nicer that way. 🤓


gurdiga:
@rtfeldman Ping! 🙃

Does this thing work? If there is anything to fix here, I’d be glad to. 🤓 Or if there is a [CONTRIBUTING.md](https://github.com/blog/1184-contributing-guidelines) somewhere that I can read about the further steps with this. 👷‍♂️

rtfeldman:
Thanks for working on this!

Initial thoughts:

1. `fill` and `visible` should be overloaded to work with this, so `pointerEvents fill` and `pointerEvents visible` can work instead of `pointerEventsFill` and `pointerEventsVisible`
2. We need to support `pointerEvents all`
3. I'd like the `{-| -}` doc comments to be replaced with docs that link to `pointerEvents` so there's a direct trail people can use to figure out what it's intended to be used with.

gurdiga:
@rtfeldman Replaced `pointerEventsVisible` with `pointerEvents visible` — changed the definition of `visible` to fit by following the example of `scroll`. If that doesn’t work well, I’m open to ideas and guidance. 🤓 

As for the `pointerEvents fill` and `pointerEvents all`, it’s my understanding that we can’t have `fill` and `all` **as values** because they already exist to designate **properties**. :neckbeard: So besides `pointerEventsFill` I have also added `pointerEventsAll`.

Added descriptions and links for the values, and also listed the new things in the Properties, Other values, and Misc `@docs` sections. 📜

Anything else? 🤓

gurdiga:
@rtfeldman The `AppVeyor` part of build seems to have failed because of a failed download: 😶
```
Error: The following HTTP request failed.
<https://github.com/eeue56/elm-lazy-list/zipball/1.0.0/>
eeue56-elm-lazy-list-b3637ce/tests/elm-package.json: openBinaryFile: does not exist (No such file or directory)
```
Checked that URL and it seems to be working now, so I think this could have been a temporary issue. :neckbeard: 

Looked around its UI but couldn’t find a restart button, is it because I don’t have enough privileges? 🤔

gurdiga:
@rtfeldman Ping again! 🙃

(Sorry fo the impatience! If there are other ways to progress with this other than poking at you, I’m open to any of them. 🤓)

tolgap:
@gurdiga If you want to immediately use this, you could clone `elm-css` into your project like this:

```
/
  elm-package.json
  /src
    Main.elm
  /vendor
    /elm-css
```

and change the `source-directories` array inside `elm-package.json` to include `vendor/elm-css` as well. You can then use git checkout to switch to any feature branch you want 👍 . You could even add your own `elm-css` fork as a different `origin` to switch between your fork and master.

rtfeldman:
Thanks @gurdiga!

I've merged this, but because it's a `MAJOR` change, I want to batch it with https://github.com/rtfeldman/elm-css/issues/375 - which is a ways off from being finished.

I appreciate your patience on this. ❤️

PR #379, opened 2018-01-12
Fix bug with borderBottomWidth functions using incorrect Animatable

robertbasden:
Fix bug with borderBottomWidth functions using BorderBottomRightRadius instead of BorderBottomWidth


robertbasden:
Failed due to elm-format issue, will re-submit once sorted

PR #380, opened 2018-01-13
Fix bug with borderBottomWidth functions using incorrect Animatable

robertbasden:
borderBottomWidth functions were using BorderBottomRightRadius instead of BorderBottomWidth


rtfeldman:
Thanks @robertbasden!

PR #389, opened 2018-03-27
Add withAttribute

lukewestby:
This combinator allows for extending a selector with additional [attribute selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors)

```elm
import Css exposing (display, none)
import Css.Foreign exposing (withAttribute, typeSelector)

typeSelector "custom-element"
  [ withAttribute "data-custom-attribute" [ display none ]
  , withAttribute "data-custom-attribute=true" [ display none ]
  ]
```

results in
```css
custom-element[data-custom-attribute] {
  display:none;
}

custom-element[data-custom-attribute=true] {
  display:none;
}
```

This is just meant as a primitive for building inline attribute selectors. Type-safe versions of the various attribute value comparison operators can be added later or separately.

### DIFF
--- a/src/Css/Foreign.elm
+++ b/src/Css/Foreign.elm
@@ -83,6 +83,7 @@ module Css.Foreign
         , typeSelector
         , ul
         , video
+        , withAttribute
         , withClass
         )
 
@@ -107,7 +108,7 @@ third-party DOM nodes!
 
 # Combinators
 
-@docs children, descendants, adjacentSiblings, generalSiblings, each, withClass
+@docs children, descendants, adjacentSiblings, generalSiblings, each, withAttribute, withClass
 
 
 # Basic elements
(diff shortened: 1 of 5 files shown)


src/Css/Foreign.elm
```diff
@@ -364,6 +365,12 @@ withClass class =
     Preprocess.ExtendSelector (Structure.ClassSelector (identifierToString "" class))
 
 
+{-| -}
```
rtfeldman: @lukewestby mind adding an example so folks can see how to use this?

lukewestby: @rtfeldman can i make a separate pull request that adds examples to all the combinators?

rtfeldman: Absolutely! That would be 😻

rtfeldman: To make the second PR easier, I'll merge this but hold off publishing until after that PR lands!

PR #394, opened 2018-04-01
fix docs typo form snake to kebap case

ChristophP:
The docs have a small error here
> Snake-case CSS names become camelCase names in elm-css.

because CSS names are kebap case (dashes) not snake case (underscores).


rtfeldman:
Thanks @ChristophP!

Issue #392, opened 2018-03-30
Phantom Types TODO List


rtfeldman:
This issue will help coordinate efforts on https://github.com/rtfeldman/elm-css/issues/375 (the [`phantom-types` branch](https://github.com/rtfeldman/elm-css/tree/phantom-types))

If you'd like to pitch in, awesome! Here are some guidelines on how to contribute: https://github.com/rtfeldman/elm-css/wiki/Phantom-Types:-Contributing

If you'd like to grab one (or a related set) of these, comment below and I'll add your name! This way we can avoid stepping on each others' toes.

### Available to convert

1. `backgroundImage`
1. `backgroundPosition`
1. `backgroundPosition2`
1. `backgroundRepeat`
1. `backgroundRepeat2`
1. `backgroundSize`
1. `backgroundSize2`
1. `blink`
1. `border`
1. `border2`
1. `border3`
1. `borderBlockEnd`
1. `borderBlockEnd2`
1. `borderBlockEnd3`
1. `borderBlockEndColor`
1. `borderBlockEndStyle`
1. `borderBlockStart`
1. `borderBlockStart2`
1. `borderBlockStart3`
1. `borderBlockStartColor`
1. `borderBlockStartStyle`
1. `borderBottom`
1. `borderBottom2`
1. `borderBottom3`
1. `borderBottomColor`
1. `borderBottomLeftRadius`
1. `borderBottomLeftRadius2`
1. `borderBottomRightRadius`
1. `borderBottomRightRadius2`
1. `borderBottomStyle`
1. `borderBottomWidth`
1. `borderCollapse`
1. `borderColor`
1. `borderColor2`
1. `borderColor3`
1. `borderColor4`
1. `borderImageOutset`
1. `borderImageOutset2`
1. `borderImageOutset3`
1. `borderImageOutset4`
1. `borderImageWidth`
1. `borderImageWidth2`
1. `borderImageWidth3`
1. `borderImageWidth4`
1. `borderInlineEnd`
1. `borderInlineEnd2`
1. `borderInlineEnd3`
1. `borderInlineEndColor`
1. `borderInlineEndStyle`
1. `borderInlineEndWidth`
1. `borderInlineStart`
1. `borderInlineStart2`
1. `borderInlineStart3`
1. `borderInlineStartColor`
1. `borderInlineStartStyle`
1. `borderLeft`
1. `borderLeft2`
1. `borderLeft3`
1. `borderLeftColor`
1. `borderLeftStyle`
[...]

PR #395, opened 2018-04-03
phantom types: linear gradient

BrianHicks:
I started to do `backgroundImage` but couldn't implement all the values without linear gradients. So this is done.

P.S. It was *so* tempting to name `toCorner` `to2` 😜

#392


rtfeldman:
Lovely, thank you @BrianHicks! 😻

PR #396, opened 2018-04-03
Phantom types background image

BrianHicks:
This builds on #395, adding `backgroundImage`. The commits for this depend on linear gradients to reach full coverage as is in master, and start at b7a4333.

8b1250d is new coverage of the CSS spec. I added it as I was thinking about it (since you can't have multiple `none`s) but feel free to cherry-pick/drop. 😄

progress for #392

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -3222,7 +3222,7 @@ linearGradient :
     Value { colorStop : Supported }
     -> Value { colorStop : Supported }
     -> List (Value { colorStop : Supported })
-    -> Style
+    -> Value { provides | linearGradient : Supported }
 linearGradient (Value firstStop) (Value secondStop) moreStops =
     let
         peeledStops =
@@ -3231,7 +3231,7 @@ linearGradient (Value firstStop) (Value secondStop) moreStops =
         stops =
             String.join "," (firstStop :: secondStop :: peeledStops)
     in
-    AppendProperty ("linear-gradient(" ++ stops ++ ")")
+    Value ("linear-gradient(" ++ stops ++ ")")
 
 
 {-| Sets [`linear-gradient`](https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient), providing an angle.
@@ -3254,7 +3254,7 @@ linearGradient2 :
     -> Value { colorStop : Supported }
     -> Value { colorStop : Supported }
     -> List (Value { colorStop : Supported })
-    -> Style
+    -> Value { provides | linearGradient : Supported }
 linearGradient2 (Value angle) (Value firstStop) (Value secondStop) moreStops =
     let
         peeledStops =
@@ -3263,7 +3263,7 @@ linearGradient2 (Value angle) (Value firstStop) (Value secondStop) moreStops =
         stops =
             String.join "," (firstStop :: secondStop :: peeledStops)
     in
-    AppendProperty ("linear-gradient(" ++ angle ++ "," ++ stops ++ ")")
+    Value ("linear-gradient(" ++ angle ++ "," ++ stops ++ ")")
 
 
 {-| Provides a stop for a [gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient).
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -37,7 +37,7 @@ module Css
         , bold
         , bolder
         , borderBox
-        , bottom
+        , bottom_
         , boxShadow
         , cell
         , center
@@ -218,7 +218,7 @@ module Css
         , titlingCaps
         , to
         , toCorner
-        , top
+        , top_
         , turn
         , unicase
         , unsafeCenter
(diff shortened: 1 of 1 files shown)


src/Css.elm
```diff
@@ -3171,6 +3209,200 @@ backgroundOrigins firstValue values =
...
+{-| Sets [`linear-gradient`](https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient), providing an angle.
+
+    linearGradient2 (deg 90) (stop red) (stop blue) []
+
+    linearGradient2 (top left) (stop red) (stop blue) []
+
+See also [`linearGradient`](#linearGradient) if you don't need to set the angle.
+
+-}
+linearGradient2 :
```
rtfeldman: I think this one can be called `linearGradient` and we can drop the other one. Offering a separate function where you don't specify the angle does not seem worth it.

src/Css.elm
```diff
@@ -3171,6 +3209,200 @@ backgroundOrigins firstValue values =
...
+    Value
+        { to : Supported
+        , deg : Supported
+        , grad : Supported
+        , rad : Supported
+        , turn : Supported
```
rtfeldman: We should add `zero : Supported` to this list (since it is supported!)

src/Css.elm
```diff
@@ -3171,6 +3209,200 @@ backgroundOrigins firstValue values =
...
+        peeledStops =
+            List.map (\(Value stop) -> stop) moreStops
+
+        stops =
+            String.join "," (firstStop :: secondStop :: peeledStops)
+    in
+    Value ("linear-gradient(" ++ angle ++ "," ++ stops ++ ")")
+
+
+{-| Provides a stop for a [gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient).
```
rtfeldman: Any reason not to link to [the CSS Tricks one](https://css-tricks.com/snippets/css/css-linear-gradient/) here?

src/Css.elm
```diff
@@ -3171,6 +3209,200 @@ backgroundOrigins firstValue values =
...
+    -> Style
+backgroundImage (Value value) =
+    AppendProperty ("background-image:" ++ value)
+
+
+{-| Sets [`background-image`](https://css-tricks.com/almanac/properties/b/background-image/) for multiple images.
+
+    backgroundImages
+        (linearGradient (stop red) (stop blue))
+        [ url "http://www.example.com/chicken.jpg" ]
```
rtfeldman: 😋 🐔

src/Css.elm
```diff
@@ -3171,6 +3209,200 @@ backgroundOrigins firstValue values =
     AppendProperty ("background-origin:" ++ str)
 
 
+{-| Sets [`background-image`](https://css-tricks.com/almanac/properties/b/background-image/).
+
+    backgroundImage (url "http://www.example.com/chicken.jpg")
+
+    backgroundImage (linearGradient (stop red) (stop blue))
+
+See also [`backgroundImage`](#backgroundImage) if you need multiple images.
```
rtfeldman: Typo? I'm guessing this text (and link!) should be `backgroundImages` plural.

src/Css.elm
```diff
@@ -1417,6 +1443,18 @@ right =
     Value "right"
 
 
+{-| -}
+top : Value { provides | top : Supported }
+top =
+    Value "top"
```
rtfeldman: I think these will need to be `top_`, `bottom_`, `left_`, and `right_` because there are already functions with those names (e.g. `top (px 20)` used with absolute positioning.) 😅 

I'd change both the value names as well as making these `top_ : Supported` so the docs and error messages and value names are all consistent!

I'd also add docs to them which mention the other ones - e.g. the docs for `top_` could say "The `top` value used in color stops. This value is called `top_` instead of `top` because [`top` is a function](#top) already."

src/Css.elm
```diff
@@ -3171,6 +3209,200 @@ backgroundOrigins firstValue values =
...
+        List
+            (Value
+                { url : Supported
+                , linearGradient : Supported
+                }
+            )
+    -> Style
+backgroundImages (Value first) rest =
+    let
+        peeled =
```
rtfeldman: I haven't seen the term "peeled" used for this operation but it totally makes sense. 😄

src/Css.elm
```diff
@@ -3171,6 +3209,200 @@ backgroundOrigins firstValue values =
...
+            , vh : Supported
+            , vw : Supported
+            , vmin : Supported
+            , vmax : Supported
+            , mm : Supported
+            , cm : Supported
+            , inches : Supported
+            , pt : Supported
+            , pc : Supported
+            , pct : Supported
```
rtfeldman: Also `zero : Supported`!

src/Css.elm
```diff
@@ -3375,3 +3607,55 @@ inset =
...
+
+    grad 38.8
+
+-}
+grad : Float -> Value { provides | grad : Supported }
+grad gradians =
+    Value (toString gradians ++ "grad")
+
+
+{-| A [`rad` angle](https://developer.mozilla.org/en-US/docs/Web/CSS/angle)
```
rtfeldman: This angle is totally rad. 😁

src/Css.elm
BrianHicks: I'll change left_ and right_ too, since they'll have the same problem and they're already `Value`s

src/Css.elm
BrianHicks: no reason whatsoever other than I didn't find it!

BrianHicks: although it can also be used for `radial-gradient`... but that's probably fine

BrianHicks:
all fixed and ready for another round.

rtfeldman:
Love it! Thanks so much @BrianHicks!

PR #399, opened 2018-04-12
Phantom types -- textTransform

sturgman:
I was missing the `@docs` lines, but I think this should be complete.

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -386,6 +386,12 @@ Multiple CSS properties use these values.
 
 @docs auto, none
 
+
+## Text Transform
+
+@docs textTransform
+@docs capitalize, uppercase, lowercase, fullWidth
+
 -}
 
 import Css.Preprocess as Preprocess exposing (Style(..))
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -36,6 +36,7 @@ module Css
         , bolder
         , borderBox
         , boxShadow
+        , capitalize
         , cell
         , center
         , ch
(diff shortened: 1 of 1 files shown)


src/Css.elm
```diff
@@ -3375,3 +3386,50 @@ inset =
...
+    textTransform capitalize
+    textTransform uppercase
+
+-}
+textTransform :
+    Value
+        { capitalize : Supported
+        , uppercase : Supported
+        , lowercase : Supported
+        , full-width : Supported
```
rtfeldman: Close! This should be `fullWidth : Supported` to match the name on the `elm-css` side.

The idea is that if someone gets an error message because they passed the wrong type to `textTransform`, the error message should list all the values that are permitted - such as `fullWidth`!

src/Css.elm
```diff
@@ -3375,3 +3386,50 @@ inset =
...
+        , inherit : Supported
+        , unset : Supported
+        }
+    -> Style
+textTransform (Value str) =
+    AppendProperty ("text-transform:" ++ str)
+
+{-| A `capitalize` [text transform](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform).
+-}
+capitalize : Value { provides | capitalize : Supported }
```
rtfeldman: Can `capitalize` (and the other values - `uppercase`, `lowercase`, and `fullWidth`) get code examples like the one `textTransform` has?

Even something as basic as showing `textTransform capitalize` can show people a valid way to use this value.

sturgman:
Hey! I added examples to the values and fixed full-width -> fullWidth.

sturgman:
Did not know whether to merge the upstream changes that you made to the repo or not. It seemed like it would result in less work for your so I gave it a shot. Thanks for your patience as I sort out my first open source (tiny) contribution.

PR #400, opened 2018-04-16
Rewrite overflow and overflowWrap using phantom types approach

juanedi:
Restored properties:
  - `overflow`
  - `overflowX`
  - `overflowY` 
  - `overflowWrap`


rtfeldman:
Awesome, thanks @juanedi!

PR #401, opened 2018-04-18
Implemented textRendering and its values

sturgman:
Implemented `textRendering`.


rtfeldman:
Wow, that was fast! Looks great @sturgman! 😻

PR #403, opened 2018-04-18
Implemented textOrientation and its values

sturgman:
Hey! Implemented `textOrientation` and its values: `mixed`, `sideways`, and `upright`. There was no suitable link to css-tricks' almanac so I used MDN.

PR #404, opened 2018-04-18
Phantom types: table properties

mattjbray:
Add properties relating to tables: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Table

Also adds a couple of overlooked exports from #403.


sturgman:
Thanks for catching those!

rtfeldman:
Sweet! Thank you @mattjbray! :heart_eyes_cat:

PR #410, opened 2018-04-22
make travis work for the phantom-types branch

BrianHicks:
tests aren't supposed to work for phantom-types until the end, so why don't we disable 'em until that integration work happens? As an external contributor, I'd love to know if my branch is actually passing, and as the author I'm sure you'd appreciate folks taking care of errors on their own 😄


rtfeldman:
Makes sense to me!

PR #408, opened 2018-04-21
Add phantom types for position and z-index

tolgap:
Various positioning related properties and values added.

@rtfeldman I wanted to ask you about the docs changes I made to `fixed`. `fixed` is a value used by a few other properties, but its docs only mentioned `background-attachment`. I simply added them all into the docs for `fixed`. LMK if you'd like it solved in a different way 👍 .

PR #411, opened 2018-04-23
Phantom types: background-repeat

BrianHicks:
This adds `background-repeat`

Progress on my bits of #392

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -4035,9 +4035,18 @@ backgroundRepeat2 :
         , unset : Supported
         , inherit : Supported
         }
+    ->
+        Value
+            { repeat : Supported
+            , space : Supported
+            , round : Supported
+            , initial : Supported
+            , unset : Supported
+            , inherit : Supported
+            }
     -> Style
-backgroundRepeat2 (Value repeat) =
-    AppendProperty ("background-repeat:" ++ repeat)
+backgroundRepeat2 (Value horiz) (Value vert) =
+    AppendProperty ("background-repeat:" ++ horiz ++ " " ++ vert)
 
 
 {-| [Repeat a background](#backgroundRepeat).
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -32,6 +32,7 @@ module Css
         , backgroundPosition
         , backgroundPosition2
         , backgroundPosition4
+        , backgroundRepeat
         , baseline
         , batch
         , before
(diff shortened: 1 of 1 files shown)


src/Css.elm
```diff
@@ -4014,6 +4015,29 @@ backgroundRepeat (Value repeat) =
...
+backgroundRepeat2 :
+    Value
+        { repeat : Supported
+        , space : Supported
+        , round : Supported
+        , initial : Supported
+        , unset : Supported
+        }
+    -> Style
+backgroundRepeat2 (Value repeat) =
```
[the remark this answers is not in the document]
BrianHicks: haaaaaaaaaaaaa yeah

src/Css.elm
```diff
@@ -3978,6 +3987,110 @@ backgroundPosition4 (Value horiz) (Value horizAmount) (Value vert) (Value vertAm
...
+
+{-| [Repeat a background](#backgroundRepeat).
+-}
+repeat : Value { provides | repeat : Supported }
+repeat =
+    Value "repeat"
+
+
+{-| [Repeat a background](#backgroundRepeat) side to side.
+-}
```
rtfeldman: On these `Value` docs, can we get links to https://developer.mozilla.org/en-US/docs/Web/CSS/background-repeat#Values and mention that this compiles to the `"repeat-x"` value in CSS? (Similar to [this](https://github.com/rtfeldman/elm-css/pull/408/files#diff-854293bbb03f363ae4c16745b0af2a75R1200), for example.)

Code examples would also be great, even if they're copy/pasted from the `background` docs. 😄

src/Css.elm
```diff
@@ -3978,6 +3987,110 @@ backgroundPosition4 (Value horiz) (Value horizAmount) (Value vert) (Value vertAm
...
+
+
+{-| Sets [`background-repeat`](https://css-tricks.com/almanac/properties/b/background-repeat/) along the horizontal axis, then the vertical axis.
+
+    backgroundRepeat2 repeat space
+
+    backgroundRepeat2 space round
+
+If you only need to set one value for both, see
+[`backgroundRepeat`](#backgroundRepeat) instead.
```
rtfeldman: Love the explanation and linking between `backgroundRepeat` and `backgroundRepeat2`! 😍

BrianHicks:
did!

rtfeldman:
Wonderful! Thank you @BrianHicks! 😻

PR #412, opened 2018-04-25
Phantom border

owanturist:
Hi @rtfeldman! I'm glad to show `border` properties as a part of #392 here. Please take a look.

## Checklist

- [x] Each function returning `Style` accepts a **closed record** of `Supported` fields, which always includes `inherit`, `initial`, and `unset` because all CSS properties support those three values! For example, `borderFoo : Value { foo : Supported, bar : Supported, inherit : Supported, initial : Supported, unset : Supported } -> Style`

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -50,6 +50,10 @@ module Css
         , borderColor4
         , borderSpacing
         , borderSpacing2
+        , borderWidth
+        , borderWidth2
+        , borderWidth3
+        , borderWidth4
         , bottom
         , bottom_
         , boxShadow
@@ -379,7 +383,7 @@ All CSS properties can have the values `unset`, `initial`, and `inherit`.
 
 ## Border
 
-@docs border, border2, border3 ,borderColor ,borderColor2 ,borderColor3 ,borderColor4
+@docs border, border2, border3, borderColor, borderColor2, borderColor3, borderColor4, borderWidth, borderWidth2, borderWidth3, borderWidth4
 
 
 ## Border Width
@@ -4370,6 +4374,9 @@ border :
         , thin : Supported
         , medium : Supported
         , thick : Supported
+        , inherit : Supported
+        , initial : Supported
+        , unset : Supported
         }
     -> Style
 border (Value width) =
@@ -4403,6 +4410,9 @@ border2 :
         , thin : Supported
         , medium : Supported
         , thick : Supported
+        , inherit : Supported
+        , initial : Supported
+        , unset : Supported
         }
     ->
         Value
@@ -4452,6 +4462,9 @@ border3 :
         , thin : Supported
         , medium : Supported
         , thick : Supported
+        , inherit : Supported
+        , initial : Supported
+        , unset : Supported
         }
     ->
         Value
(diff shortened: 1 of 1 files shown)


src/Css.elm
```diff
@@ -4317,66 +4382,1913 @@ listStyle3 (Value val1) (Value val2) (Value val3) =
...
+
+    border (px 1)
+    border2 (px 1) solid
+    border3 (px 1) solid (hex "#f00")
 
-{-| A `wavy` [text decoration style](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-style#Values).
 -}
-wavy : Value { provides | wavy : Supported }
-wavy =
-    Value "wavy"
```
owanturist: I've removed `wavy` value, because it's related with `text-decoration-style` which hasn't implemented yet and I thought it can confuse a contributor who will implement it. Should we keep changes or come `wavy` back?

src/Css.elm
```diff
@@ -369,9 +409,36 @@ All CSS properties can have the values `unset`, `initial`, and `inherit`.
...
 
-@docs wavy, dotted, dashed, solid, double, groove, ridge, inset, outset
+@docs borderStyle, borderStyle2, borderStyle3, borderStyle4, borderTopStyle, borderRightStyle, borderBottomStyle, borderLeftStyle
+
+@docs dotted, dashed, solid, double, groove, ridge, inset, outset
+
+
+## Border Color
+
+@docs borderColor, borderColor2, borderColor3, borderColor4, borderTopColor, borderRightColor, borderBottomColor, borderLeftColor
```
owanturist: I'm little bit worry about the docs structure but this kind of approach the most logical and clear by my opinion.

src/Css.elm
```diff
@@ -598,8 +665,8 @@ type Value supports
 
 {-| A type used to specify which properties and which values work together.
 -}
-type Supported
-    = Supported
+type alias Supported =
+    Never
```
owanturist: Little proposition for making phantom types more phantom :) Like never before.

owanturist: Actually there are more variants:

1. If we don't want that user will describe phantom types of `Value` with both `Never` and `Supported`
	```elm
	type Supported
		= Supported Never
	```
2. We may not use any special type, just empty tuple. Error messages keep same.
	```elm
	dotted : Value { provides | dotted : () }
	dotted =
	    Value "dotted"
	```

rtfeldman: A nice feature of having `Supported` be a union type that doesn't expose its constructor is that other modules can't instantiate it. This limits how other modules can use it, which in this case is desirable!

On the other hand, `Never` and `()` can be instantiated outside this module (by using `Basics.never` and `()`, respectively), which means we're potentially giving up guarantees around how these values can be constructed, without necessarily knowing what those guarantees are are or why we're giving them up. 😅

To make a long story short, I definitely want to leave this as-is! 😄

src/Css.elm
rtfeldman: This grouping seems fine to me! 👍

src/Css.elm
rtfeldman: I think it's fine to leave it. It already happens that implementors run into things that have already been implemented (e.g. `left` is reused in several places), so I think this should be fine!

owanturist:
@rtfeldman issues have been completed

src/Css.elm
```diff
@@ -4317,104 +4396,1951 @@ listStyle3 (Value val1) (Value val2) (Value val3) =
...
+groove : Value { provides | groove : Supported }
+groove =
+    Value "groove"
+
+
+{-| The `ridge` [`border-style`](<https://css-tricks.com/almanac/properties/b/border/#article-header-id-0> value.
+
+    borderStyle ridge
+
+Similar to `groove`, but reverses the color values in a way that makes the element appear raised.
```
rtfeldman: Love the concise descriptions on these! 😻

PR #413, opened 2018-04-26
Phantom border radius

owanturist:
Hi @rtfeldman! The next bulk of properties as a part of #392 are here 😀 Please take a look.

## Checklist

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -6448,6 +6448,634 @@ outset =
 
 
 
+{- BORDER RADIUS -}
+
+
+{-| Sets [`border-radius`](https://css-tricks.com/almanac/properties/b/border-radius/) property.
+
+    borderRadius  (em 4)
+    borderRadius2 (em 4) (px 2)
+    borderRadius3 (em 4) (px 2) (pct 5)
+    borderRadius4 (em 4) (px 2) (pct 5) (px 3)
+
+-}
+borderRadius :
+    Value
+        { ch : Supported
+        , cm : Supported
+        , em : Supported
+        , ex : Supported
+        , inches : Supported
+        , mm : Supported
+        , pc : Supported
+        , pt : Supported
+        , pct : Supported
+        , px : Supported
+        , rem : Supported
+        , vh : Supported
+        , vmax : Supported
+        , vmin : Supported
+        , vw : Supported
+        , zero : Supported
+        , inherit : Supported
+        , initial : Supported
+        , unset : Supported
+        }
+    -> Style
+borderRadius (Value radius) =
+    AppendProperty ("border-radius:" ++ radius)
+
+
+{-| Sets [`border-radius`](https://css-tricks.com/almanac/properties/b/border-radius/) property.
+
+    borderRadius  (em 4)
+    borderRadius2 (em 4) (px 2)
+    borderRadius3 (em 4) (px 2) (pct 5)
+    borderRadius4 (em 4) (px 2) (pct 5) (px 3)
+
+-}
+borderRadius2 :
+    Value
+        { ch : Supported
+        , cm : Supported
+        , em : Supported
+        , ex : Supported
+        , inches : Supported
+        , mm : Supported
+        , pc : Supported
+        , pt : Supported
+        , pct : Supported
+        , px : Supported
+        , rem : Supported
+        , vh : Supported
+        , vmax : Supported
+        , vmin : Supported
+        , vw : Supported
+        , zero : Supported
+        , inherit : Supported
+        , initial : Supported
+        , unset : Supported
+        }
+    ->
+        Value
+            { ch : Supported
+            , cm : Supported
+            , em : Supported
+            , ex : Supported
+            , inches : Supported
+            , mm : Supported
+            , pc : Supported
+            , pt : Supported
+            , pct : Supported
+            , px : Supported
+            , rem : Supported
+            , vh : Supported
+            , vmax : Supported
+            , vmin : Supported
+            , vw : Supported
+            , zero : Supported
+            , inherit : Supported
+            , initial : Supported
+            , unset : Supported
+            }
+    -> Style
+borderRadius2 (Value radiusTopLeftAndBottomRight) (Value radiusTopRightAndBottomLeft) =
+    AppendProperty ("border-radius:" ++ radiusTopLeftAndBottomRight ++ " " ++ radiusTopRightAndBottomLeft)
+
+
+{-| Sets [`border-radius`](https://css-tricks.com/almanac/properties/b/border-radius/) property.
+
+    borderRadius  (em 4)
+    borderRadius2 (em 4) (px 2)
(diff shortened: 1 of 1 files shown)


src/Css.elm
```diff
@@ -6448,6 +6465,685 @@ outset =
...
+{-| Sets [`border-radius`](https://css-tricks.com/almanac/properties/b/border-radius/) property.
+
+    borderRadius  (em 4)
+    borderRadius2 (em 4) (px 2)
+    borderRadius3 (em 4) (px 2) (pct 5)
+    borderRadius4 (em 4) (px 2) (pct 5) (px 3)
+
+**Note:** An implementation of shorthand property with supporting of elipse shape looks ugly,
+so if you wanna make somehting like this `border-radius: 4px 3px 6px / 2px 4px` by `elm-css`
+you should create a batch of `Style`:
```
owanturist: I didn't know how to say about the implementation problem more polite.. and also my English is not so cool as should be and I need your help as a native speaker 😅

src/Css.elm
```diff
@@ -6448,6 +6465,685 @@ outset =
...
+
+    [ borderTopLeftRadius     (px 4) (px 2)
+    , borderTopRightRadius    (px 3) (px 4)
+    , borderBottomRightRadius (px 6) (px 2)
+    , borderBottomLeftRadius  (px 3) (px 4)
+    ]
+
+**Note:** As with any shorthand property, individual sub-properties cannot inherit,
+such as in `borderRadius2 (px 1) inherit`, which would partially override existing definitions.
+Instead, the individual longhand properties have to be used.
```
owanturist: There is a limitation which I had never known (and I'm ashamed about it). I'm not sure that we should encapsulate things like this but the point is easy to do. So I've removed `inherit: Supported` for each of `borderRadius{2,3,4}` and `border{TopLeft,TopRight,BotomRight,BottomLeft}Radius2` and keep for same single-argument properties. What do you thing about it? Should we handle the limitation here or developers should do it self?

src/Css.elm
rtfeldman: Wow, I never knew about `/` in `border-radius`!

I wonder if we should consider supporting this. We could do that by adding something like `, ellipticalRadius : Supported` to each of the `Value` arguments in `borderRadius`, then introducing this:
```
ellipticalRadius :
    Value { px : Supported, em : Supported, ... }
    -> Value { px : Supported, em : Supported, ... }
    -> Value { provides | ellipticalRadius : Supported }
```

so then you could call:

```
borderRadius (ellipticalRadius (px 10) (px 30))
```

thoughts @owanturist?

owanturist: The main problem is specific scheme of this property. I'd like to see the scheme like this:

```css
border-radius: 1px/2px 3px 4px/5px 6px/7px;
```

Where `1px`, `3px`, `4px` and `6px` are general values (like in `border-radius: 1px 3px 4px 6px`) and rest (`2px`, `5px` and `7px`) are addition values.

But it's a wrong signature and correct equivalent of example is:

```css
border-radius: 1px 3px 4px 6px / 2px 0 5px 7px;
```

So, you can see that supporting of `/` in `borderRadius` is not trivial, because all of them are valid:

```css
border-radius: 1px / 2px;
border-radius: 1px / 2px 3px;
border-radius: 1px / 2px 3px 4px;
border-radius: 1px / 2px 3px 4px 5px;

border-radius: 1px 2px / 3px;
border-radius: 1px 2px / 3px 4px;
border-radius: 1px 2px / 3px 4px 5px;
border-radius: 1px 2px / 3px 4px 5px 6px;

border-radius: 1px 2px 3px / 4px;
border-radius: 1px 2px 3px / 4px 5px;
border-radius: 1px 2px 3px / 4px 5px 6px;
border-radius: 1px 2px 3px / 4px 5px 6px 7px;

/* end so on... */
```

Well, if we wanna implement `ellipticalCorner` (which can be reused in `font` property if the rule would looks like `border-radius: 1px/2px 3px 4px/5px 6px/7px`) we could implement `borderRadius` from single argument to nine (`borderRadius9 (px 1) (px 2) (px 3) (px 4) ellipticalCorner (px 5) (px 6) (px 7) (px 8)` is most long) or pass tuple of list to `ellipticalCorner` but in this case we loose type checking. But even in `borderRadius9` we can't guarantee that developer pass `ellipticalCorner` in right place (`borderRadius9 (px 1) ellipticalCorner (px 2) (px 3) (px 4) (px 5) (px 6) (px 7) (px 8)` for example) so we should implement the property `borderRadius4elliptical4` for correct type checking:

```elm
borderRadius4elliptical4 (px 1) (px 2) (px 3) (px 4) (px 5) (px 6) (px 7) (px 8)
```

@rtfeldman sorry for this long story, but yesterday I broke my head while I was thinking about it 🤯

src/Css.elm
```diff
@@ -6448,6 +6472,1022 @@ outset =
...
+        , vmax : Supported
+        , vmin : Supported
+        , vw : Supported
+        , zero : Supported
+        , initial : Supported
+        , unset : Supported
+        }
+    -> Value { properties | elliptical : Supported }
+elliptical (Value radius) =
+    Value (" / " ++ radius)
```
owanturist: @rtfeldman I've improved your idea and it look like thank

src/Css.elm
rtfeldman: Oh wow, and it's true of `initial` and `unset` too! 😮 

Great catch @owanturist! I think we should definitely change the other properties to reflect this, and I updated the contributing docs to note this: https://github.com/rtfeldman/elm-css/wiki/Phantom-Types:-Contributing

owanturist: I didn't check it for `initial` and `unset` but you're right! I think we should reflect this.

owanturist:
@rtfeldman I've gotten rid from `unset` and `initial` in multivalues properties. Could you take a look please?

src/Css.elm
```diff
@@ -6448,6 +6472,912 @@ outset =
...
+            , pt : Supported
+            , pct : Supported
+            , px : Supported
+            , rem : Supported
+            , vh : Supported
+            , vmax : Supported
+            , vmin : Supported
+            , vw : Supported
+            , zero : Supported
+            , elliptical : Supported
```
rtfeldman: Hm, looking at how this turned out, I think it's gonna be pretty confusing to people why `elliptical` is only supported on the final argument. (This is correct, since the spec says you only get one `/` maximum, but I think it will surprise people who are unfamiliar with this.)

I'm leaning towards saying we shouldn't support the `/` after all. [The docs suggest](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius) it's just shorthand for other properties, so it appears there's nothing you can't express a different way if we just didn't support `elliptical`.

That's sounding simpler at this point...what do you think @owanturist?

rtfeldman:
Super great! Thank you for all the iteration on this one @owanturist - I'm really happy with the final result! 😻

PR #416, opened 2018-04-30
Add float

brianvanburken:
## Contribution Checklist

- [x] Each function returning `Style` accepts a **closed record** of `Supported` fields, which always includes `inherit`, `initial`, and `unset` because all CSS properties support those three values! For example, `borderFoo : Value { foo : Supported, bar : Supported, inherit : Supported, initial : Supported, unset : Supported } -> Style`

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -136,6 +136,7 @@ module Css
         , fixed
         , flexEnd
         , flexStart
+        , float
         , fontFamilies
         , fontFamily
         , fontSize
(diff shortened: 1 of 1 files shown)


src/Css.elm
```diff
@@ -635,6 +636,11 @@ Multiple CSS properties use these values.
 @docs whiteSpace
 @docs pre, preWrap, preLine, nowrap
 
+
+## Float
+
+@docs float
```
brianvanburken: @rtfeldman do I need to add a line containing the values `left_`, `right_`, and `none` here? Since those are already defined in with other CSS properties

rtfeldman: Nah, I'd say leave them where they are! 👍

PR #418, opened 2018-05-01
Add visibility

brianvanburken:
## Contribution Checklist


rtfeldman:
Lovely, thank you @brianvanburken!

PR #419, opened 2018-05-01
Phantom types: fix wordbreak

BrianHicks:
`wordBreak` currently has the wrong signature in phantom-types :cry: This fixes it. I *think* it's right! But always good to have a second set of eyes on it 😄

Also adds a Travis check for type signatures so future issues on phantom-types should fail the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,5 +38,8 @@ script:
   - elm-make --output=/dev/null --yes
   - cd ../readme-example
   - elm-make --output=/dev/null --yes
-  # - cd ..
+  - cd ..
+  # this line is here to catch type errors until phantom-types is merged into
+  # master, at which point it can be replaced by the test script again. :)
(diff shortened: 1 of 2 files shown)


brianvanburken:
As far as I can see `word-break` is not a value but a property so it shouldn't have `provides`. Source: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break. I see that I did forget to add `-> Style` in #417. I apologize for the mistake! :(

BrianHicks:
1. yep! I missed that in another branch too. Awhoops! Once the test suite works again it will catch these. Thanks for the catch. <3
2. you're fine, these things happen. Just needed a little tweak, and I've added a check for that to CI. Thanks for writing it in the first place!

rtfeldman:
Looks good, thanks @BrianHicks!

And yeah, totally don't sweat it @brianvanburken 😄 - thanks for the contribution!

PR #420, opened 2018-05-01
Styled.styled supports tesk9/accessible-html-with-css by being more general

michaelglass:
(paired with @avh4)

Previously:

when using `Html.Styled.styled`, with `tesk9/accessible-html-with-css`, `styled` would coerce all subtypes into `Html Never`

This PR generalizes `Html.Styled.styled` such that it can still return `Html msg`


rtfeldman:
Makes sense, thanks @michaelglass!

PR #421, opened 2018-05-01
Phantom types: background-size

BrianHicks:
This adds `backgroundSize` and `backgroundSize2` for #392. Commits for this start at 48733cd, and I'm happy to rebase onto phantom-types when/if #419 is merged.

I was going to add `backgroundSizes` to go along with `backgroundImages`, but I can't figure out a good way of doing the types. Essentially, we need to be able to have all these properties:

```css
/* valid */
background-size: cover;
background-size: 100px;
background-size: 100px, cover;
background-size: 100px 20px, cover;

/* invalid */
backgrouns-size: cover cover;
background-size: 100px 20px, cover cover;
```

---

- [x] If a function returning `Style` takes a single `Value`, that `Value` should always support `inherit`, `initial`, and `unset` because all CSS properties support those three values!
- [x] If a function returning `Style` takes **more than one** `Value`, however, then **none** of its arguments should support `inherit`, `initial`, or `unset`, because these can never be used in conjunction with other values!

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -4348,9 +4348,6 @@ backgroundSize2 :
             , vmin : Supported
             , vmax : Supported
             , auto : Supported
-            , inherit : Supported
-            , initial : Supported
-            , unset : Supported
         }
     ->
         Value
@@ -4373,9 +4370,6 @@ backgroundSize2 :
                 , vmin : Supported
                 , vmax : Supported
                 , auto : Supported
-                , inherit : Supported
-                , initial : Supported
-                , unset : Supported
             }
     -> Style
 backgroundSize2 (Value width) (Value height) =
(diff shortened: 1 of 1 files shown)


src/Css.elm
```diff
@@ -4231,6 +4237,123 @@ round =
...
+            , pt : Supported
+            , ch : Supported
+            , em : Supported
+            , ex : Supported
+            , rem : Supported
+            , vh : Supported
+            , vw : Supported
+            , vmin : Supported
+            , vmax : Supported
+            , auto : Supported
```
rtfeldman: Nice, thanks for not including `inherit` et al 😄

src/Css.elm
```diff
@@ -4231,6 +4237,123 @@ round =
...
+            }
+    -> Style
+backgroundSize2 (Value width) (Value height) =
+    AppendProperty ("background-size:" ++ width ++ " " ++ height)
+
+
+{-| Used in [`backgroundSize`](#backgroundSize) to always show the whole
```
rtfeldman: Can we get CSS Tricks (or, failing that, MDN) links for `contain` and `cover`?

rtfeldman:
Fantastic, thank you @BrianHicks!

PR #436, opened 2018-06-10
Phantom types: border-image-outset and border-image-width

Arkham:
This adds `borderImageOutset`, `borderImageOutset2`, `borderImageOutset3`, `borderImageOutset4`, `borderImageWidth`,  `borderImageWidth2`,  `borderImageWidth3` and  `borderImageWidth4` for https://github.com/rtfeldman/elm-css/issues/392.

It also changes the type of pure numeric value that can be passed to `zIndex` from `num` to `int`.

## Contribution Checklist

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -66,6 +66,10 @@ module Css
         , borderImageOutset2
         , borderImageOutset3
         , borderImageOutset4
+        , borderImageWidth
+        , borderImageWidth2
+        , borderImageWidth3
+        , borderImageWidth4
         , borderLeft
         , borderLeft2
         , borderLeft3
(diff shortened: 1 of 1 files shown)


Arkham:
Not sure why CI is failing, locally elm-validate doesn't produce any warning.

src/Css.elm
```diff
@@ -1482,15 +1497,15 @@ sticky =
...
+    zIndex (int 10)
     zIndex auto
 
 **NOTE:** Z-index is not as simple as it looks! Make sure to read about [stacking contexts](https://css-tricks.com/css-stacking-contexts/) if you're not already familiar with them.
 
 -}
 zIndex :
     Value
-        { num : Supported
+        { int : Supported
```
rtfeldman: 😍

PR #440, opened 2018-06-17
[0.19] Improve Lazy API, add more lazyN functions

### LINKED ISSUE #364: Idea about Lazy
I have an idea how to improve Lazy API and still avoid #348:

```elm
lazy : (a -> Html msg) -> a -> Html msg
lazy fn arg =
    VirtualDom.lazy2 lazyHelp fn arg
        |> Unstyled


lazyHelp : (a -> Html msg) -> a -> VirtualDom.Node msg
lazyHelp fn arg =
    fn arg
        |> toUnstyled
```

VirtualDom compares arguments by reference in JS so we can safely pass `fn` to it. The problem is that `lazy` is implemented with the usage of `VirtualDom.lazy2` and `lazy2` will then use `VirtualDom.lazy3`... and it seems that there is no way to implement `lazy3` in which case it must use a different API. I don't know if it's worth it (or maybe there is a solution for `lazy3`?), just wanted to share this.

And here is the proof that it actually works https://ellie-app.com/kxHM94kNHa1/2. You should see a delay when you press the button, which is caused by an expensive computation in the view, after removing `lazy` on line 30.

edkv:
Closes #364.

### DIFF
--- a/src/Html/Styled/Lazy.elm
+++ b/src/Html/Styled/Lazy.elm
@@ -1,34 +1,22 @@
-module Html.Styled.Lazy exposing (lazy, lazy2, lazy3)
+module Html.Styled.Lazy
+    exposing
+        ( lazy
+        , lazy2
+        , lazy3
+        , lazy4
+        , lazy5
+        , lazy6
+        , lazy7
+        )
 
-{-| A not-quite-drop-in replacement for the `Html.Lazy` module from the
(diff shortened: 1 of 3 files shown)


rtfeldman:
Cool, thanks @edkv!

Have you verified that the `lazyHelp` approach works? (It should, just want to double check.)

PR #448, opened 2018-07-08
Phantom types: Add transform

brianvanburken:
## Properties
- transform
- transformN

## Values
- perspective
- rotate
- rotate3d
- rotateX
- rotateY
- rotateZ
- scale
- scale2
- scale3d
- scaleX
- scaleY
- scaleZ
- skew
- skew2
- skewX
- skewY
- translate
- translate2
- translate3d
- translateX
- translateY
- translateZ

# Contributing
- [x] When accepting a numeric `Value` (e.g. a length like `px`, an angle like `deg`, or a unitless number like `int` or `num`), *always* include `zero : Supported` as well as `calc : Supported`!


src/Css.elm
```diff
@@ -11417,3 +11478,740 @@ columnRule3 :
...
+transform (Value val) =
+    AppendProperty ("transform:" ++ val)
+
+
+{-| Sets [`transform`](https://css-tricks.com/almanac/properties/t/transform/).
+
+    transformN [ (translate (px 12)), (scale 2), (skew (deg 20)) ]
+
+-}
+transformN :
```
brianvanburken: @rtfeldman do you know a way to enforce one or more in a list? This functions allows for zero or more right now

stoeffel: @brianvanburken you can use a tuple for that `(a, List a)`.

rtfeldman: `( a, List a )` works, although I think stylistically elsewhere in `elm-css` we've used `a -> List a -> ...` (as in, first argument is the first element in the list, second argument is the rest of it) so we should probably stick with that to be consistent!

src/Css.elm
```diff
@@ -4199,7 +4260,7 @@ fontVariantNumeric4 val1 val2 val3 val4 =
 
 maybeValToString : Maybe (Value a) -> Maybe String
 maybeValToString =
-    Maybe.map (\(Value val) -> val)
+    Maybe.map unpackValue
```
rtfeldman: 👍 makes sense!

src/Css.elm
rtfeldman: That said, @brianvanburken in cases where there's a useful and obvious value for `[]` to output, I like to do that and mention it in the docs.

So in this case I'd add something to the documentation about "if you pass `[]` you'll get `transform: none`" and implement that behavior. I think that'll be more useful than requiring a minimum of 1 element.

rtfeldman: One other note on this one: I think it's fine to rename `transformN` to `transform` and have that be the only option.

That will make it easier to discover, and adding an extra `[` and `]` around the argument shouldn't be too much bother. 😄

rtfeldman:
Thanks @brianvanburken! Looks awesome! 😻

PR #451, opened 2018-07-17
Phantom Type: Add missing color to box-shadow output

brianvanburken:
Found this little bug when I was working on the `textShadow` implementation. Color was not present in the output.


rtfeldman:
Ah, nice catch! Thanks @brianvanburken

PR #452, opened 2018-07-26
Phantom types: add most missing flexbox functions/values

rlefevre:
Hi, first PR here so feedback is welcome.

**Important notes / feedback likely needed**
1. I replaced `displayFlex` by `display flex_` as it seems that `underscore` won the namespace war for now. Tell me if I'm wrong.
2. I included a fix to "close" the "open" record parameter of `alignItems` and `all` (in passing) as they seemed wrong (their value record was parameterized on `accepts`). But I may have misunderstood something so tell me if I should remove this modification.
3. I did not  include yet overflow and baseline alignment for `justifyContent` and `alignContent` as they are still in a draft specification and are not well supported by all browsers. However `alignItems` already implements them partially (overflow alignment is limited to `safe center` or `unsafe center` but `center | start | end | self-start | self-end | flex-start | flex-end` should also be valid). So I need feedback to know what to do.
4. I also did not include intrinsic sizing (`min-content`, `max-content`, ...) for `flexBasis` for the same reasons. However they are listed in the [phantom-types issue](https://github.com/rtfeldman/elm-css/issues/392), so I'm not really sure what we want.

**Other less important notes**
* I added a 3rd level of documentation to try to have a nice flexbox section. If this is not wanted, I can flatten out the flexbox documentation to 2 levels as the rest of the document.
* I renamed `flexFlow1` (in the listed function to implement) to `flexFlow` as it seemed more consistent with other multiple values functions
* I struggled with the functions definition order. Maybe something should be added in the guidelines about that or a final pass done once all types are implemented.
* They are some inconsistency between comments about the use of values terminated by an `underscore` in the source. I choose what seemed the best but a pass to harmonize them is most likely needed at the end.

Thank you very much for this library.
Regards

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -197,6 +197,8 @@ module Css
         , flexBasis
         , flexDirection
         , flexEnd
+        , flexFlow
+        , flexFlow2
         , flexGrow
         , flexShrink
         , flexStart
@@ -709,6 +711,11 @@ See this [complete guide](https://css-tricks.com/snippets/css/a-guide-to-flexbox
 ### Flexbox Wrapping
 
 @docs flexWrap, nowrap, wrap, wrapReverse
+
+
+### Flexbox Shorthands
+
+@docs flexFlow, flexFlow2
 
 
 ## Font Size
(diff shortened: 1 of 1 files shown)


src/Css.elm
```diff
@@ -3597,6 +3708,408 @@ alignSelf (Value val) =
...
+flexWrap (Value val) =
+    AppendProperty ("flex-wrap:" ++ val)
+
+
+{-| The `nowrap` [`flex-wrap` value](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-wrap#Values).
+
+    flexWrap noWrap
+
+-}
+noWrap : Value { provides | noWrap : Supported }
```
brianvanburken: Awesome work! `nowrap` is all lowercase as a value. Should we be consistent with the values and make this also all lowercase?

rlefevre: Actually, `nowrap` already existed for `white-space`, so I removed `noWrap` and added an example in `nowrap`.

Thank you!

PR #453, opened 2018-07-30
Phantom types aliases proposal

rlefevre:
When writing the [flexbox phantom types](https://github.com/rtfeldman/elm-css/pull/452), I noticed that using a value that can include a [CSS length](https://developer.mozilla.org/en-US/docs/Web/CSS/length) was quite error prone.

 It required a copy/paste from another similar function or from the spec and double checking each supported value in the record, without [forgetting](https://github.com/rtfeldman/elm-css/blob/32f1b4d0ff836652cb7a6d9ccdbfd600dc70270b/src/Css.elm#L11087) to add `zero` and `calc`. Also if a new unit is added one day (as `rem` or others before), it would be quite difficult to be sure to have updated all functions that support a `length` as they are not easily identified (there are also some inconsistencies in the order of the record values for example). 

At last, making single `Value` supporting `inherit`, `initial` and `unset` also required some attention and could be quite easily misdone (for example look at the [spurious `inherit` in the last `Value` of `borderWidth3` current version](https://github.com/rtfeldman/elm-css/blob/32f1b4d0ff836652cb7a6d9ccdbfd600dc70270b/src/Css.elm#L6955)).

This proposal aims at improving this.

I would advise to read first each commit separately to ease the understanding of the quite large modifications.

## [1. Introduce phantom types aliases](https://github.com/rtfeldman/elm-css/pull/453/commits/e7c880c5336a6f08ec288d7e700784bcc9d1bca4)

This commit introduces phantom types aliases to improve the reliability, readability and ease of modification of the library without loosing the type check properties and error messages of the `phantom-type` branch.

Ex:
```elm
{-| A type alias used to accept a length among other values.
-}
type alias LengthSupported v =
    { v  
        | ch : Supported
        , em : Supported
        , ex : Supported
        , rem : Supported
        , vh : Supported
        , vw : Supported
        , vmin : Supported
        , vmax : Supported
        , px : Supported
        , cm : Supported
        , mm : Supported
        , inches : Supported
        , pc : Supported
        , pt : Supported
        , zero : Supported
        , calc : Supported
    }    


{-| A type alias used to accept a length.
-}
type alias Length =
    LengthSupported {}
```

* the type aliases can be combined, even with duplicate fields

Ex:
```elm
type alias BorderWidthSupported v =
    LengthSupported
        { v
            | thin : Supported
            , medium : Supported
            , thick : Supported
        }

borderLeft :
    Value
        (BorderWidthSupported
            { inherit : Supported
            , initial : Supported
            , unset : Supported
            }
        )
    -> Style
```
* errors messages are strictly identical as before (ie: very nice in this branch)

Ex:
```elm
top :
    Value
        (LengthSupported
            { pct : Supported
            , auto : Supported
            , inherit : Supported
            , initial : Supported
            , unset : Supported
            }
        )
    -> Style
[...]


main = 
    toUnstyled <|
        div
            [ css
                [ top (rgb 2 2 2)
                ]
            ]
            []
```
raises the following error:
```elm
-- TYPE MISMATCH ------------------------------------------------------ Main.elm

The argument to function `top` is causing a mismatch.

18|                   top (rgb 2 2 2)
                           ^^^^^^^^^
Function `top` is expecting the argument to be:

    Value
        { auto : ...
        , calc : ...
        , ch : ...
        , cm : ...
        , em : ...
        , ex : ...
        , inches : ...
        , inherit : ...
        , initial : ...
        , mm : ...
        , pc : ...
        , pct : ...
        , pt : ...
        , px : ...
        , rem : ...
        , unset : ...
        , vh : ...
        , vmax : ...
        , vmin : ...
        , vw : ...
        , zero : ...
        }

But it is:

    Value { provides | rgb : ... }

Hint: The record fields do not match up. One has auto, calc, ch, cm, em, ex,
inches, inherit, initial, mm, pc, pct, pt, px, rem, unset, vh, vmax, vmin, vw,
and zero. The other has rgb.
```
* the signatures might be considered less explicit, but I would say that they are closer to MDN documentation and more readable when there are a lot of supported values (like for `length`). Look at the `margin` functions signature for example:

```elm
margin :
    Value
        (LengthSupported
            { pct : Supported
            , auto : Supported
            , inherit : Supported
            , initial : Supported
            , unset : Supported
            }
        )
    -> Style

margin4 :
    Value
        (LengthSupported
            { pct : Supported
            , auto : Supported
            }
        )
    ->
        Value
            (LengthSupported
                { pct : Supported
                , auto : Supported
                }
            )
    ->
        Value
            (LengthSupported
                { pct : Supported
                , auto : Supported
                }
            )
    ->
        Value
            (LengthSupported
                { pct : Supported
                , auto : Supported
                }
            )
    -> Style

```
With `LengthSupported` being a link to its definition.

Compared to the initial version:

```elm
margin :
    Value
        { ch : Supported
        , cm : Supported
        , em : Supported
        , ex : Supported
        , inches : Supported
        , mm : Supported
        , pc : Supported
        , pt : Supported
        , pct : Supported
        , px : Supported
        , rem : Supported
        , vh : Supported
        , vmax : Supported
        , vmin : Supported
        , vw : Supported
        , zero : Supported
        , calc : Supported
        , auto : Supported
        , inherit : Supported
        , initial : Supported
        , unset : Supported
        }
    -> Style 

margin4 :
    Value
        { ch : Supported
        , cm : Supported
        , em : Supported
        , ex : Supported
        , inches : Supported
        , mm : Supported
        , pc : Supported
        , pt : Supported
        , pct : Supported
        , px : Supported
        , rem : Supported
        , vh : Supported
        , vmax : Supported
        , vmin : Supported
        , vw : Supported
        , zero : Supported
        , calc : Supported
        , auto : Supported
        }
    ->
        Value
            { ch : Supported
            , cm : Supported
            , em : Supported
            , ex : Supported
            , inches : Supported
            , mm : Supported
            , pc : Supported
            , pt : Supported
            , pct : Supported
            , px : Supported
            , rem : Supported
            , vh : Supported
            , vmax : Supported
            , vmin : Supported
            , vw : Supported
            , zero : Supported
            , calc : Supported
            , auto : Supported
            }
   ->
        Value
            { ch : Supported
            , cm : Supported
            , em : Supported
            , ex : Supported
            , inches : Supported
            , mm : Supported
            , pc : Supported
            , pt : Supported
            , pct : Supported
            , px : Supported
            , rem : Supported
            , vh : Supported
            , vmax : Supported
            , vmin : Supported
            , vw : Supported
            , zero : Supported
            , calc : Supported
            , auto : Supported
            }
    ->
        Value
            { ch : Supported
            , cm : Supported
            , em : Supported
            , ex : Supported
            , inches : Supported
            , mm : Supported
            , pc : Supported
            , pt : Supported
            , pct : Supported
            , px : Supported
            , rem : Supported
            , vh : Supported
            , vmax : Supported
            , vmin : Supported
            , vw : Supported
            , zero : Supported
            , calc : Supported
            , auto : Supported
            }

```
 

* the type aliases help to automatically support `zero` or `calc` for types that support it (like `LengthSupported` ones). For example, look at the [missing `zero` and `calc` in `strokeSize` function current version](https://github.com/rtfeldman/elm-css/blob/32f1b4d0ff836652cb7a6d9ccdbfd600dc70270b/src/Css.elm#L11087).

## [2. Introduce Stylist* types aliases](https://github.com/rtfeldman/elm-css/pull/453/commits/40181bc35f5f898f136d101bbc431cf25d924326)

For the special case of global values (`inherit`, `initial` and `unset`), we could also use a type alias like in the previous commit. However this proposal goes one step further by introducing aliases that help to guarantee the following [implementation guidelines](https://github.com/rtfeldman/elm-css/wiki/Phantom-Types:-Contributing):
> - if a function returning Style takes a single Value, that Value should always support inherit, initial, and unset 
> - If a function returning Style takes more than one Value, however, then none of its arguments should support inherit, initial, or unset

```elm
{-| A type alias for functions returning a `Style` from one `Value`. This alias
will make the `Value` support the `initial`, `inherit` and `unset` global values.
-}
type alias Stylist v =
    Value
        { v
            | initial : Supported
            , inherit : Supported
            , unset : Supported
        }
    -> Style

{-| A type alias for functions returning a `Style` from two `Value`.
-}
type alias Stylist2 v1 v2 =
    Value v1 -> Value v2 -> Style

-- then Stylist3, Stylist4, ...
```

This commit make the signatures a little less explicit (the returned value), but it is quite helpful and signatures are quickly understandable once almost every function use the Stylist* types, so I'm not sure if it is too much or not.

Ex:
```elm
margin :
    Stylist
        (LengthSupported
            { pct : Supported
            , auto : Supported
            }
        )

margin4 :
    Stylist4
        (LengthSupported
            { pct : Supported
            , auto : Supported
            }
        )
        (LengthSupported
            { pct : Supported
            , auto : Supported
            }
        )
        (LengthSupported
            { pct : Supported
            , auto : Supported
            }
        )
        (LengthSupported
            { pct : Supported
            , auto : Supported
            }
        )

borderColor : Stylist Color

borderColor4 : Stylist4 Color Color Color Color
```

## Conclusion

If you look into details, you will see that these commits already fix some small mistakes and a lot of inconsistencies, so it seems clear to me that they make errors more visible and harder to make. 

But this is only a proposal, open for discussion, that could be added partially, modified, rejected, ...
Tell me what you think.

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -1,8 +1,19 @@
 module Css
     exposing
-        ( BoxShadowConfig
+        ( Angle
+        , AngleSupported
+        , BorderStyle
+        , BorderStyleSupported
+        , BorderWidth
+        , BorderWidthSupported
+        , BoxShadowConfig
         , CalcOperation
         , Color
+        , ColorSupported
+        , Length
+        , LengthSupported
+        , ListStyle
+        , ListStyleSupported
         , Style
         , Supported
         , Value
@@ -534,6 +545,8 @@ All CSS properties can have the values `unset`, `initial`, and `inherit`.
 
 @docs all, revert
 
+@docs Length, LengthSupported, Angle, AngleSupported
+
 
 ## Numeric Units
 
@@ -547,7 +560,7 @@ All CSS properties can have the values `unset`, `initial`, and `inherit`.
 
 ## Color
 
-@docs Color, color, backgroundColor, hex, rgb, rgba, hsl, hsla
+@docs Color, ColorSupported, color, backgroundColor, hex, rgb, rgba, hsl, hsla
 
 
 ## Pseudo-Classes
@@ -592,6 +605,8 @@ All CSS properties can have the values `unset`, `initial`, and `inherit`.
 
 
 ## Border
+
+@docs BorderWidth, BorderWidthSupported, BorderStyle, BorderStyleSupported
 
 @docs border, border2, border3
 
@@ -729,6 +744,7 @@ All CSS properties can have the values `unset`, `initial`, and `inherit`.
 
 ## List Style Type
 
+@docs ListStyle, ListStyleSupported
 @docs listStyle, listStyle2, listStyle3
 @docs arabicIndic, armenian, bengali, cjkEarthlyBranch, cjkHeavenlyStem, devanagari, georgian, gujarati, gurmukhi, kannada, khmer, lao, malayalam, myanmar, oriya, telugu, thai
 
@@ -983,6 +999,119 @@ type Value supports
 -}
 type Supported
     = Supported
+
+
+{-| A type alias used to accept a [length](https://developer.mozilla.org/en-US/docs/Web/CSS/length)
+among other values.
+-}
+type alias LengthSupported v =
+    { v
+        | ch : Supported
+        , em : Supported
+        , ex : Supported
+        , rem : Supported
+        , vh : Supported
+        , vw : Supported
+        , vmin : Supported
+        , vmax : Supported
+        , px : Supported
+        , cm : Supported
+        , mm : Supported
+        , inches : Supported
+        , pc : Supported
+        , pt : Supported
+        , zero : Supported
+        , calc : Supported
+    }
+
+
+{-| A type alias used to accept a [length](https://developer.mozilla.org/en-US/docs/Web/CSS/length).
+-}
+type alias Length =
+    LengthSupported {}
+
+
+{-| A type alias used to accept a [color](https://developer.mozilla.org/en-US/docs/Web/CSS/color)
+among other values.
+-}
+type alias ColorSupported v =
+    { v
+        | rgb : Supported
+        , rgba : Supported
+        , hsl : Supported
+        , hsla : Supported
+        , hex : Supported
+        , transparent : Supported
+        , currentColor : Supported
+    }
+
+
+{-| A type alias used to accept a [color](https://developer.mozilla.org/en-US/docs/Web/CSS/color).
+-}
+type alias Color =
+    ColorSupported {}
+
+
+{-| A type alias used to accept a [border-style](https://developer.mozilla.org/en-US/docs/Web/CSS/border-style#Values)
+among other values.
+-}
+type alias BorderStyleSupported v =
+    { v
+        | none : Supported
+        , hidden : Supported
+        , dotted : Supported
+        , dashed : Supported
+        , solid : Supported
+        , double : Supported
+        , groove : Supported
+        , ridge : Supported
+        , inset : Supported
+        , outset : Supported
+    }
+
+
+{-| A type alias used to accept a [border-style](https://developer.mozilla.org/en-US/docs/Web/CSS/border-style#Values).
+-}
+type alias BorderStyle =
+    BorderStyleSupported {}
+
+
+{-| A type alias used to accept a [border-width](https://developer.mozilla.org/en-US/docs/Web/CSS/border-width#Values)
+among other values.
+-}
+type alias BorderWidthSupported v =
+    LengthSupported
+        { v
+            | thin : Supported
+            , medium : Supported
+            , thick : Supported
+        }
+
+
+{-| A type alias used to accept a [border-width](https://developer.mozilla.org/en-US/docs/Web/CSS/border-width#Values).
+-}
+type alias BorderWidth =
+    BorderWidthSupported {}
+
+
+{-| A type alias used to accept an [angle](https://developer.mozilla.org/en-US/docs/Web/CSS/angle)
+among other values.
+-}
+type alias AngleSupported v =
+    { v
+        | deg : Supported
+        , grad : Supported
+        , rad : Supported
+        , turn : Supported
+        , zero : Supported
+        , calc : Supported
+    }
+
+
+{-| A type alias used to accept an [angle](https://developer.mozilla.org/en-US/docs/Web/CSS/angle).
+-}
+type alias Angle =
+    AngleSupported {}
 
 
 
@@ -1319,31 +1448,6 @@ breakWord =
 -- COLORS --
 
 
-{-| A color created with [`hex`](#hex), [`rgb`](#rgb), [`rgba`](#rgba),
-[`hsl`](#hsl), or [`hsla`](#hsla). `Color` can be used to annotate values
-returned by any of those functions.
-
-    limeGreen : Css.Color
-    limeGreen =
-        rgb 0 100 44
-
-    rebeccaPurple : Css.Color
-    rebeccaPurple =
-        hex "#663399"
-
--}
-type alias Color =
-    Value
-        { rgb : Supported
-        , rgba : Supported
-        , hsl : Supported
-        , hsla : Supported
-        , hex : Supported
-        , transparent : Supported
-        , currentColor : Supported
-        }
-
-
 {-| Sets [`color`](https://css-tricks.com/almanac/properties/c/color/).
 
     color (hex "#60b5cc")
@@ -1353,17 +1457,12 @@ type alias Color =
 -}
 color :
     Value
-        { rgb : Supported
-        , rgba : Supported
-        , hsl : Supported
-        , hsla : Supported
-        , hex : Supported
-        , transparent : Supported
-        , currentColor : Supported
-        , inherit : Supported
-        , initial : Supported
-        , unset : Supported
-        }
+        (ColorSupported
+            { inherit : Supported
+            , initial : Supported
+            , unset : Supported
+            }
+        )
     -> Style
 color (Value val) =
     AppendProperty ("color:" ++ val)
@@ -1378,17 +1477,12 @@ color (Value val) =
 -}
 backgroundColor :
     Value
-        { rgb : Supported
-        , rgba : Supported
-        , hsl : Supported
-        , hsla : Supported
-        , hex : Supported
-        , transparent : Supported
-        , currentColor : Supported
-        , inherit : Supported
-        , initial : Supported
-        , unset : Supported
-        }
+        (ColorSupported
+            { inherit : Supported
+            , initial : Supported
+            , unset : Supported
+            }
+        )
     -> Style
 backgroundColor (Value val) =
     AppendProperty ("background-color:" ++ val)
@@ -1529,28 +1623,14 @@ for example in `vertical-align: top`, use [`top_`](#top_) instead of this.
 -}
 top :
     Value
-        { px : Supported
-        , cm : Supported
-        , mm : Supported
-        , inches : Supported
-        , pc : Supported
-        , pct : Supported
-        , pt : Supported
-        , ch : Supported
-        , em : Supported
-        , ex : Supported
-        , rem : Supported
-        , vh : Supported
-        , vw : Supported
-        , vmin : Supported
-        , vmax : Supported
-        , auto : Supported
-        , zero : Supported
-        , calc : Supported
-        , inherit : Supported
-        , initial : Supported
-        , unset : Supported
-        }
+        (LengthSupported
+            { pct : Supported
+            , auto : Supported
+            , inherit : Supported
+            , initial : Supported
+            , unset : Supported
+            }
+        )
     -> Style
 top (Value val) =
     AppendProperty ("top:" ++ val)
@@ -1569,28 +1649,14 @@ for example in `vertical-align: bottom`, use [`bottom_`](#bottom_) instead of th
 -}
 bottom :
     Value
-        { px : Supported
-        , cm : Supported
-        , mm : Supported
-        , inches : Supported
-        , pc : Supported
-        , pct : Supported
-        , pt : Supported
-        , ch : Supported
-        , em : Supported
-        , ex : Supported
-        , rem : Supported
-        , vh : Supported
-        , vw : Supported
-        , vmin : Supported
-        , vmax : Supported
-        , auto : Supported
-        , zero : Supported
-        , calc : Supported
-        , inherit : Supported
-        , initial : Supported
-        , unset : Supported
-        }
+        (LengthSupported
+            { pct : Supported
+            , auto : Supported
+            , inherit : Supported
+            , initial : Supported
+            , unset : Supported
+            }
+        )
     -> Style
 bottom (Value val) =
     AppendProperty ("bottom:" ++ val)
@@ -1609,28 +1675,14 @@ for example in `float: left`, use [`left_`](#left_) instead of this.
 -}
 left :
     Value
-        { px : Supported
-        , cm : Supported
-        , mm : Supported
-        , inches : Supported
-        , pc : Supported
-        , pct : Supported
-        , pt : Supported
-        , ch : Supported
-        , em : Supported
-        , ex : Supported
-        , rem : Supported
-        , vh : Supported
-        , vw : Supported
-        , vmin : Supported
-        , vmax : Supported
-        , auto : Supported
-        , zero : Supported
-        , calc : Supported
-        , inherit : Supported
-        , initial : Supported
-        , unset : Supported
-        }
+        (LengthSupported
+            { pct : Supported
+            , auto : Supported
+            , inherit : Supported
+            , initial : Supported
+            , unset : Supported
+            }
+        )
     -> Style
 left (Value val) =
     AppendProperty ("left:" ++ val)
@@ -1649,28 +1701,14 @@ for example in `float: right`, use [`right_`](#right_) instead of this.
 -}
 right :
     Value
-        { px : Supported
-        , cm : Supported
-        , mm : Supported
-        , inches : Supported
-        , pc : Supported
-        , pct : Supported
-        , pt : Supported
-        , ch : Supported
-        , em : Supported
-        , ex : Supported
-        , rem : Supported
-        , vh : Supported
-        , vw : Supported
-        , vmin : Supported
-        , vmax : Supported
-        , auto : Supported
-        , zero : Supported
-        , calc : Supported
-        , inherit : Supported
-        , initial : Supported
-        , unset : Supported
-        }
+        (LengthSupported
+            { pct : Supported
+            , auto : Supported
+            , inherit : Supported
+            , initial : Supported
+            , unset : Supported
+            }
+        )
     -> Style
 right (Value val) =
     AppendProperty ("right:" ++ val)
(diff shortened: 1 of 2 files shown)


rtfeldman:
Intriguing! I'm surprised to learn the error messages look the same way with this proposal. That's cool.

I have some questions:

1. Do the error messages still look the same way in 0.19 with and without the type aliases?
1. [In the `Stylist` commit](https://github.com/rtfeldman/elm-css/pull/453/commits/40181bc35f5f898f136d101bbc431cf25d924326) the `ListStyleSupported` type alias is only ever used to define the `ListStyle` type alias, which suggests we don't need to expose both. Is that true of all the `___Supported` aliases? Can they all be dropped in the `Stylist` world?
1. Whenever I've used type aliases for functions I've ended up regretting it, so while I like the goal of the `Stylist` idea, I'm not on board with the design of having it alias functions. However, it's conceivable that we could rename `Value` to `ExplicitValue` and then introduce `type alias Value supported = Value { supported | initial : Supported , inherit : Supported , unset : Supported }`. This would mean (for example) that `borderColor` and `borderColor2` would become `borderColor : Value Color -> Style` and `borderColor2 : ExplicitValue Color -> ExplicitValue Color -> Style`. What do you think of that design?

rlefevre:
1. I don't think we have a `phantom-types` branch compatible with 0.19, so I made the following test that should be close enough to `elm-css`:

```elm
module Main exposing (main)

import Html exposing (Html, text)

type Value provides
    = Value String

{-| A type used to specify which properties and which values work together.
-}
type Supported
    = Supported

top :
    Value
        { ch : Supported
        , em : Supported
        , ex : Supported
        , rem : Supported
        , vh : Supported
        , vw : Supported
        , vmin : Supported
        , vmax : Supported
        , px : Supported
        , cm : Supported
        , mm : Supported
        , inches : Supported
        , pc : Supported
        , pt : Supported
        , zero : Supported
        , calc : Supported
        }
    -> String
top (Value val) =
    val 

rgba : Int -> Int -> Int -> Float -> Value { provides | rgba : Supported }
rgba r g b a = 
    Value "rgba"

main : Html msg 
main =
    text (top (rgba 0 0 0 0))
```
The error reported is
```
-- TYPE MISMATCH ------------------------------------------------------ Main.elm

The 1st argument to `top` is not what I expect:

47|     text (top (rgba 0 0 0 0))
                   ^^^^^^^^^^^^
This `rgba` call produces:

    Value
        { a
            | calc : Supported
            , ch : Supported
            , cm : Supported
            , em : Supported
            , ex : Supported
            , inches : Supported
            , mm : Supported
            , pc : Supported
            , pt : Supported
            , px : Supported
            , rem : Supported
            , rgba : Supported
            , vh : Supported
            , vmax : Supported
            , vmin : Supported
            , vw : Supported
            , zero : Supported
        }

But `top` needs the 1st argument to be:

    Value
        { calc : Supported
        , ch : Supported
        , cm : Supported
        , em : Supported
        , ex : Supported
        , inches : Supported
        , mm : Supported
        , pc : Supported
        , pt : Supported
        , px : Supported
        , rem : Supported
        , vh : Supported
        , vmax : Supported
        , vmin : Supported
        , vw : Supported
        , zero : Supported
        }

Hint: Seems like a record field typo. Maybe rgba should be rem?

Hint: Can more type annotations be added? Type annotations always help me give
more specific messages, and I think they could help a lot in this case!
```
So now, elm adds fields supported by `top` in the value returned by `rgba`. But with the coloring (the `rgba` field is actually in yellow in the record), I would say that the message is still quite good. Without colors, it may be more difficult to see the difference between the two types (the first `Hint` helps also a little).

If we add some type aliases:
```elm
top :
    Value
        (LengthSupported
            { pct : Supported
            }
        )
    -> String
top (Value val) =
    val 
```
we get:
```
-- TYPE MISMATCH ------------------------------------------------------ Main.elm

The 1st argument to `top` is not what I expect:

64|     text (top (rgba 0 0 0 0))
                   ^^^^^^^^^^^^
This `rgba` call produces:

    Value
        { a
            | calc : Supported
            , ch : Supported
            , cm : Supported
            , em : Supported
            , ex : Supported
            , inches : Supported
            , mm : Supported
            , pc : Supported
            , pct : Supported
            , pt : Supported
            , px : Supported
            , rem : Supported
            , rgba : Supported
            , vh : Supported
            , vmax : Supported
            , vmin : Supported
            , vw : Supported
            , zero : Supported
        }

But `top` needs the 1st argument to be:

    Value (LengthSupported { pct : Supported })

Hint: Seems like a record field typo. Maybe rgba should be rem?

Hint: Can more type annotations be added? Type annotations always help me give
more specific messages, and I think they could help a lot in this case!
```
(with `rgba` highlighted in yellow in the record)

If the type alias is "closed":
```elm
top : Value Length -> String
top (Value val) =
    val 
```
we get:
```
-- TYPE MISMATCH ------------------------------------------------------ Main.elm

The 1st argument to `top` is not what I expect:

58|     text (top (rgba 0 0 0 0))
                   ^^^^^^^^^^^^
This `rgba` call produces:

    Value
        { a
            | calc : Supported
            , ch : Supported
            , cm : Supported
            , em : Supported
            , ex : Supported
            , inches : Supported
            , mm : Supported
            , pc : Supported
            , pt : Supported
            , px : Supported
            , rem : Supported
            , rgba : Supported
            , vh : Supported
            , vmax : Supported
            , vmin : Supported
            , vw : Supported
            , zero : Supported
        }

But `top` needs the 1st argument to be:

    Value Length

Hint: Seems like a record field typo. Maybe rgba should be rem?
```
(with `rgba` highlighted in yellow in the record)

So it seems that elm 0.19 does not "unroll" anymore type aliases in those error messages.

I would say that all of those error messages are slightly less explicit than 0.18 ones, but not as bad as the master branch. (Of course we are talking here only about the specific case of `elm-css`, I have no doubt that error messages in 0.19 have improved globally).

Do you think that the changes in 0.19 error messages compromise the added value of phantom types aliases?

2. Not really. The `Stylist*` aliases handle the global values `inherit`, `initial` and `unset`, so it is more likely that a closed type alias (like `ListStyle` instead of `ListStyleSupported`) will be enough, but this is not a rule, for example there are a lot of functions that use `Stylist (LenghtSupported { ... } )` like:
```elm
top :
    Stylist
        (LengthSupported
            { pct : Supported
            , auto : Supported
            }
        )
```

3. 
> Whenever I've used type aliases for functions I've ended up regretting it

I'm not really surprised but this is still interesting. Why have you regretted it usually? Also, for example what could be regretted in using the `Comparator` alias in `NoRedInk/elm-compare` ?
```elm
type alias Comparator a = 
    a -> a -> Order
```
(I can say that one drawback is that it prevented me to find the library when searching on https://klaftertief.github.io/elm-search/ but this could most likely be improved in the tool)

Anyway, until now, I have never met a case where using a function alias brings a real benefit. However in this case, it does. Your `ExplicitValue` suggestion is what I meant by 
> For the special case of global values (inherit, initial and unset), we could also use a type alias like in the previous commit.

but if `Stylist` must be used for all functions taking only `Value` arguments, it is impossible to use `Stylist` instead of `Stylist2` (they do not have the same number of arguments), however if I understand correctly your suggestion, it is possible to use a `Value` instead of an `ExplicitValue`, so this is not the same level of protection.

This is why I wanted to submit this `Stylist` proposal. Because it brings some real value at the cost of legibility/explicitness, it deserved a discussion. But to be honest, I usually don't like functions aliases either, so this is really a choice between the value added and the drawbacks. 

Your suggestion avoids the issues of a function alias at the cost of less guaranties, which is as much reasonable if not more, I don't know. This is really your call.

rlefevre:
Thinking more about it, there might be a solution without a function alias that is as strong, something like:

```elm
type alias Value supported =
    Value String


type alias Stylist supported =
    Value
        { supported
            | initial : Supported
            , inherit : Supported
            , unset : Supported
        }


type alias Stylist2 supported1 supported2 =
    Value2 String String


borderColor : Stylist Color -> Style
borderColor (Value val) =
   ...


borderColor2 : Stylist2 Color Color -> Style
borderColor2 (Value2 val1 val2) =
  ...
```

PS: I am still interested in the function alias drawbacks from your point of view.

rlefevre:
1. I don't think that we have a `phantom-types` branch compatible with 0.19, so I made the following test that should be close enough to `elm-css`:

```elm
module Main exposing (main)

import Html exposing (Html, text)

type Value provides
    = Value String

{-| A type used to specify which properties and which values work together.
-}
type Supported
    = Supported

top :
    Value
        { ch : Supported
        , em : Supported
        , ex : Supported
        , rem : Supported
        , vh : Supported
        , vw : Supported
        , vmin : Supported
        , vmax : Supported
        , px : Supported
        , cm : Supported
        , mm : Supported
        , inches : Supported
        , pc : Supported
        , pt : Supported
        , zero : Supported
        , calc : Supported
        }
    -> String
top (Value val) =
    val 

rgba : Int -> Int -> Int -> Float -> Value { provides | rgba : Supported }
rgba r g b a = 
    Value "rgba"

main : Html msg 
main =
    text (top (rgba 0 0 0 0))
```
The error reported is
```
-- TYPE MISMATCH ------------------------------------------------------ Main.elm

The 1st argument to `top` is not what I expect:

47|     text (top (rgba 0 0 0 0))
                   ^^^^^^^^^^^^
This `rgba` call produces:

    Value
        { a
            | calc : Supported
            , ch : Supported
            , cm : Supported
            , em : Supported
            , ex : Supported
            , inches : Supported
            , mm : Supported
            , pc : Supported
            , pt : Supported
            , px : Supported
            , rem : Supported
            , rgba : Supported
            , vh : Supported
            , vmax : Supported
            , vmin : Supported
            , vw : Supported
            , zero : Supported
        }

But `top` needs the 1st argument to be:

    Value
        { calc : Supported
        , ch : Supported
        , cm : Supported
        , em : Supported
        , ex : Supported
        , inches : Supported
        , mm : Supported
        , pc : Supported
        , pt : Supported
        , px : Supported
        , rem : Supported
        , vh : Supported
        , vmax : Supported
        , vmin : Supported
        , vw : Supported
        , zero : Supported
        }

Hint: Seems like a record field typo. Maybe rgba should be rem?

Hint: Can more type annotations be added? Type annotations always help me give
more specific messages, and I think they could help a lot in this case!
```
So now, elm adds fields supported by `top` in the value returned by `rgba`. But with the coloring (the `rgba` field is actually in yellow in the record), I would say that the message is still quite good. Without colors, it may be more difficult to see the difference between the two types (the first `Hint` helps also a little, the second one is unexpected as I don't think that any type signature is missing).

If we add some type aliases:
```elm
top :
    Value
        (LengthSupported
            { pct : Supported
            }
        )
    -> String
top (Value val) =
    val 
```
we get:
```
-- TYPE MISMATCH ------------------------------------------------------ Main.elm

The 1st argument to `top` is not what I expect:

64|     text (top (rgba 0 0 0 0))
                   ^^^^^^^^^^^^
This `rgba` call produces:

    Value
        { a
            | calc : Supported
            , ch : Supported
            , cm : Supported
            , em : Supported
            , ex : Supported
            , inches : Supported
            , mm : Supported
            , pc : Supported
            , pct : Supported
            , pt : Supported
            , px : Supported
            , rem : Supported
            , rgba : Supported
            , vh : Supported
            , vmax : Supported
            , vmin : Supported
            , vw : Supported
            , zero : Supported
        }

But `top` needs the 1st argument to be:

    Value (LengthSupported { pct : Supported })

Hint: Seems like a record field typo. Maybe rgba should be rem?

Hint: Can more type annotations be added? Type annotations always help me give
more specific messages, and I think they could help a lot in this case!
```
(with `rgba` highlighted in yellow in the record)

If the type alias is "closed":
```elm
top : Value Length -> String
top (Value val) =
    val 
```
we get:
```
-- TYPE MISMATCH ------------------------------------------------------ Main.elm

The 1st argument to `top` is not what I expect:

58|     text (top (rgba 0 0 0 0))
                   ^^^^^^^^^^^^
This `rgba` call produces:

    Value
        { a
            | calc : Supported
            , ch : Supported
            , cm : Supported
            , em : Supported
            , ex : Supported
            , inches : Supported
            , mm : Supported
            , pc : Supported
            , pt : Supported
            , px : Supported
            , rem : Supported
            , rgba : Supported
            , vh : Supported
            , vmax : Supported
            , vmin : Supported
            , vw : Supported
            , zero : Supported
        }

But `top` needs the 1st argument to be:

    Value Length

Hint: Seems like a record field typo. Maybe rgba should be rem?
```
(with `rgba` highlighted in yellow in the record)

So it seems that elm 0.19 does not "unroll" anymore type aliases in those error messages for the needed arguments. This seems a little inconsistent to me because it does it for the `rgba` function returned value.

I would say that all of those error messages are slightly less explicit than 0.18 ones, but not as bad as the master branch. (Of course we are talking here only about the specific case of `elm-css` phantom types, I have no doubt that error messages in 0.19 have improved for most cases).

Do you think that the changes in 0.19 error messages compromise the added value of phantom types aliases?

2. Not really. The `Stylist*` aliases handle the global values `inherit`, `initial` and `unset`, so it is more likely that a closed type alias (like `ListStyle` instead of `ListStyleSupported`) will be enough, but this is not a rule, for example there are a lot of functions that use `Stylist (LenghtSupported { ... } )` like:
```elm
top :
    Stylist
        (LengthSupported
            { pct : Supported
            , auto : Supported
            }
        )
```

Edit: yes, at least `LengthSupported` as `Length` is used a lot with other compatible values like `num`, `int`, `pct`, `auto`, etc. The shorthand functions that accept several different arguments also often need some `*Supported` aliases.

3. 
> Whenever I've used type aliases for functions I've ended up regretting it

I'm not really surprised but this is still interesting. Why have you regretted it usually? Also, for example what could be regretted in using the `Comparator` alias in `NoRedInk/elm-compare` ?
```elm
type alias Comparator a = 
    a -> a -> Order
```
(I can say that one drawback is that it prevented me to find the library when searching on https://klaftertief.github.io/elm-search/ but this could most likely be improved in the tool)

Anyway, until now, I have never met a case where using a function alias brings a real benefit. However in this case, it does. Your `ExplicitValue` suggestion is what I meant by 
> For the special case of global values (inherit, initial and unset), we could also use a type alias like in the previous commit.

but if `Stylist` must be used for all functions taking only `Value` arguments, it is impossible to use `Stylist` instead of `Stylist2` (they do not have the same number of arguments), however if I understand correctly your suggestion, it is possible to use a `Value` instead of an `ExplicitValue`, so this is not the same level of protection.

This is why I wanted to submit this `Stylist` proposal. Because it brings some real value at the cost of legibility/explicitness, it deserved a discussion. 

But to be honest, I usually don't like functions aliases either, so this is really a choice between the value added and the drawbacks. Your suggestion avoids the issues of a function alias at the cost of less guaranties, which is as much reasonable if not more, I don't know. This is really your call.

leifhelm:
The `Stylist` proposal crashes the elm compiler (0.19.1)!
See the small example below
```
module Test exposing (..)


type Value supporting
    = Value String


type Supported
    = Supported


type alias Stylist supported =
    Value
        { supported
            | initial : Supported
            , inherit : Supported
            , unset : Supported
        }
    -> String


consumeD : Stylist { num : Supported }
consumeD (Value val) =
    val
```
This error message is thrown by my compiler:
```
Compiling ...elm: Map.!: given key is not an element in the map
CallStack (from HasCallStack):
  error, called at libraries/containers/containers/src/Data/Map/Internal.hs:627:17 in containers-0.6.2.1:Data.Map.Internal

-- ERROR -----------------------------------------------------------------------

I ran into something that bypassed the normal error reporting process! I
extracted whatever information I could from the internal error:

>   thread blocked indefinitely in an MVar operation

These errors are usually pretty confusing, so start by asking around on one of
forums listed at https://elm-lang.org/community to see if anyone can get you
unstuck quickly.

-- REQUEST ---------------------------------------------------------------------

If you are feeling up to it, please try to get your code down to the smallest
version that still triggers this message. Ideally in a single Main.elm and
elm.json file.

From there open a NEW issue at https://github.com/elm/compiler/issues with your
reduced example pasted in directly. (Not a link to a repo or gist!) Do not worry
about if someone else saw something similar. More examples is better!

This kind of error is usually tied up in larger architectural choices that are
hard to change, so even when we have a couple good examples, it can take some
time to resolve in a solid way.elm: thread blocked indefinitely in an MVar operation
```
So I guess it is not possible with elm-0.19.1

rlefevre:
@leifhelm Yes there is a bug in 0.19.1, this requires to do it in 2 steps:

```
type alias Stylist_ supported =    
    Value    
        { supported    
            | initial : Supported    
            , inherit : Supported    
            , unset : Supported    
        }    
    
    
type alias Stylist supported =    
    Stylist_ supported -> String  
```

Not sure what would be the best name for `Stylist_` here.

leifhelm:
@rlefevre I made a mistake while testing it initially (it did crash the compiler again), but reading your suggestion I tried again and it works now.

Issue #455, opened 2018-07-31
(Suggestion) Make the `Styled.Html.styled` function lazy


Warry:
Reading through the code, I was surprised that the `styled` function didn't seemed to be lazy at all. Hence the hash is calculated every time the function is called. 

current implementation:
https://github.com/rtfeldman/elm-css/blob/2064f1523be13f8647cbba954313b736e6f8d5c3/src/Html/Styled.elm#L301-L303

my suggestion:
```elm
styled fn styles =
    let
        styleAttribute =
            Internal.css styles
    in
        \attrs children ->
            fn (styleAttribute :: attrs) children
```

This would allow to cache the computation of the hash when declaring styled nodes at the top level, like:
```elm
myNode =
    Styled.Html.styled Styled.Html.div
        [ Css.displayFlex
        ]
```

It shouldn't radically change the performances, but it's free!

rtfeldman:
Great idea @Warry!

I don't see any downsides...happy to accept a PR for this!

Warry:
Sweet, happy to help :) The PR already exists at #456

PR #456, opened 2018-07-31
Make *.Styled.styled lazy

Warry:
See the #455 issue.


Warry:
Pushed again for further simplification (removing `children`, and using currying instead), elm-019 formatting.

rtfeldman:
Lovely, thank you @Warry!

I'm going to do a release soon for the Keyframes API, so this will get included in that.

PR #462, opened 2018-08-17
Fix #461: degree/radian mismatch for `hsla` hue

### LINKED ISSUE #461: Unit mismatch for hue parameter in `Css.hsla`
`Css.hsla` takes hue in degrees (per the [CSS spec](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#hsl())), but `Color.hsl` [uses radians](http://package.elm-lang.org/packages/elm-lang/core/latest/Color#hsl).

For example, `hsla(270, 100%, 50%, 1)` should be `rgba(127, 0, 255, 1)`, a [purple](http://hslpicker.com/#7f00ff). But we actually get `rgba(255, 0, 43, 1)`, a red:
```Elm
> Css.hsl 270 1 0.5
{ value = "hsl(270, 100%, 50%)", color = Compatible, red = 255, green = 0, blue = 43, alpha = 1, warnings = [] }
    : Css.Color
```

alidaka:
I didn't see an easy way to write test for this into the existing test strategy, but I'm happy to take another look with some pointers!

I hit this bug when creating "theme" colors for my site, then trying to lighten/darken them by adjusting the `hsla` lightness. `elm-css` only preserves the rgb values, which takes a bit of conversion to round trip through `hsla`, resulting in this bug.

Thanks!
Augustus

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -2318,7 +2318,7 @@ hslaToRgba : String -> Float -> Float -> Float -> Float -> Color
 hslaToRgba value hue saturation lightness hslAlpha =
     let
         { red, green, blue, alpha } =
-            Color.hsla hue saturation lightness hslAlpha
+            Color.hsla (degreesToRadians hue) saturation lightness hslAlpha
                 |> Color.toRgb
     in
     { value = value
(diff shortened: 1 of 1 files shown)


rtfeldman:
Thanks @alidaka!

alidaka:
For sure; thank you, @rtfeldman !

PR #463, opened 2018-08-23
WIP: Upgrade to 0.19

rtfeldman:
My goal is to:

1. Get this branch ready to go (need to update README example, among other things)
2. Backport it to 0.18
3. Publish that version as the final 0.18 release
4. Publish this as the first 0.19 release, ideally as a `PATCH` change (and if not, as close to that as possible!)

PR #465, opened 2018-08-29
Final 0.18 release

rtfeldman:
**NOTE:** This does not update `examples`! I don't think that needs to block the release, so I can do that separately.

This sets us up to release 15.0.0 as both an 0.18 package (with `Html.Styled` being a drop-in replacement for 0.18's `html`) and an 0.19 package (with `Html.Styled` being a drop-in replacement for 0.19's `html`) with no CSS API differences.

Having an 0.18 and 0.19 package with the same CSS API should make it easier to incrementally upgrade!


n1k0:
I've just tested this branch by upgrading the src retrieved from #463, but this time I'm facing an error:

```
-- REDUNDANT EXPORT --------------------------------------- elm-css-0.19/Css.elm

You are trying to expose `animationName` multiple times! Once here:

7|     , animationName
         ^^^^^^^^^^^^^
And again right here:

65|     , animationName
          ^^^^^^^^^^^^^
Remove one of them and you should be all set!

elm-live:
  elm make failed! You can find more info above. Keep calm
  and take your time to fix your code. We’ll try to compile it again
  as soon as you change a file.

```

n1k0:
By removing the duplicate export and docs reference, everything seems to work perfectly :)

PR #469, opened 2018-09-03
Correct attiibute into attribute

dvberkel:
Note that this is in the documentation comments only.


rtfeldman:
Thanks for the fix @dvberkel!

PR #475, opened 2018-10-11
fix CI for phantom types

BrianHicks:
Looks like `elm-format@rc` doesn't exist anymore, and IMO it'd be better to explicit about what version of Elm we're using now that 0.19 is out (:tada:).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,24 +22,24 @@ before_install:
 
 
 install:
-  - npm install -g elm@0.18.0 elm-test@0.18.12 elm-format@rc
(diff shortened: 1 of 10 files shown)


BrianHicks:
ping @rtfeldman — does this conflict with the 0.19 upgrade? Would it make sense to upgrade the phanton-types branch to 0.19 instead of doing something like this?

rtfeldman:
Nope, this is fine!

I've been focusing more on getting `elm-test` out of beta for 0.19, since (thankfully!) the `phantom-types` branch isn't blocking anyone.

Just gave this a quick review, and it's good to merge. Thanks @BrianHicks! 🎉

PR #482, opened 2018-11-13
Add installation instructions to the README for newcomers

GabeAtWork:
I'm just starting using Elm, and it wasn't completely straightforward how to install this package from the manager.

These days, many libraries indicate installation instructions in their README (even simple stuff like `yarn add some-react-lib`).
In this PR, I'm proposing to add these instructions to the README to make it even friendlier for newcomers.


clozach:
I've had this same thought a few times, @GabeAtWork, though I suspect it could be addressed once and for all  with a change to (package.elm-lang.org's source)[https://github.com/elm/package.elm-lang.org]).

This would take more work up front, but then the issue you're bringing up here would be fixed for _every_ package, not just elm-css. Other benefits: consistency and accuracy across projects. Since installation is generally as simple as `elm install foo/bar`, solving this problem at the source would also allow for the instructions to be rendered as a simple sidebar widget with a [copy] button for convenience.

(Note that there would probably need to be some allowance made for the Core packages that are embedded in the compiler and never need to be installed.)

PR #483, opened 2018-11-14
Docs fix: class no longer accepts custom types

mthadley:
Just a quick documentation fix.

Also, I the old snippet kinda makes me laugh since it looks like we are about to do some good ol' OOP. (`class Foo`)


rtfeldman:
Great point, thanks @mthadley!

PR #492, opened 2019-01-07
Fix invalid types in example: (int 1.5) -> (num 1.5)

Herteby:


Herteby:
Btw, the reason I found this was because I was scouring the docs for how to use floats. :stuck_out_tongue: 
Might be nice to have a section listing the different unit types :slightly_smiling_face:

PR #505, opened 2019-08-19
Phantom Types: add a bunch more properties

brianvanburken:
It's been a while, but I've taken some time to add a bunch more properties to the Phantom Type branch. This should take us a lot towards releasing a new version with phantom types.

- [x] When accepting a numeric `Value` (e.g. a length like `px`, an angle like `deg`, or a unitless number like `int` or `num`), *always* include `zero : Supported` as well as `calc : Supported`!

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -112,6 +112,7 @@ module Css exposing
     , translate, translate2, translateX, translateY, translateZ, translate3d
     , opacity
     , zoom
+    , pageBreakBefore, pageBreakInside, pageBreakAfter
     , backfaceVisibility
     , bleed
     , caretColor
@@ -558,6 +559,11 @@ Multiple CSS properties use these values.
 # Viewport
 
 @docs zoom
+
+
+# Page break
+
+@docs pageBreakBefore, pageBreakInside, pageBreakAfter
 
 
 # Other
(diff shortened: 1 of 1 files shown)


src/Css.elm
```diff
@@ -12497,3 +12500,25 @@ backfaceVisibility :
...
+        , pt : Supported
+        , cm : Supported
+        , inches : Supported
+        , inherit : Supported
+        , initial : Supported
+        , unset : Supported
+        }
+    -> Style
+bleed (Value val) =
+    AppendProperty ("bleed" ++ val)
```
BrianHicks: missing a `:` here

src/Css.elm
```diff
@@ -13069,3 +13071,58 @@ lineClamp :
...
+
+    mixBlendMode hardLight
+
+    mixBlendMode softLight
+
+    mixBlendMode difference
+
+    mixBlendMode exclusion
+
+    mixBlendMode saturation
```
BrianHicks: I think it may be a little much to demonstrate examples for all 13 of these. Would you mind choosing two or so?

src/Css.elm
```diff
@@ -13525,3 +13537,1002 @@ pointerEvents :
...
+
+    scrollMargin (em 4)
+
+    scrollMargin2 (em 4) (px 2)
+
+    scrollMargin3 (em 4) (px 2) (pct 5)
+
+    scrollMargin4 (em 4) (px 2) (pct 5) (px 3)
+
+-}
```
BrianHicks: I saw an elm-css user the other day getting frustrated because the docs for margin were:

```elm
margin (px 10)
margin2 (px 10) (px 10)
margin3 (px 10) (px 10) (px 10)
margin4 (px 10) (px 10) (px 10) (px 10)
```

These have kind of the same problem—they show some allowable types but are not really useful as a reference to remember how the behavior works. I think we could do a little better here. What about documenting what each of the positions do? Does it go clockwise from the top like elsewhere in scrollMargin4? top/bottom and left/right in scrollMargin2? What's the actual behavior in the 3-value  case?

src/Css.elm
```diff
@@ -13525,3 +13537,1002 @@ pointerEvents :
...
+
+    scrollPadding (em 4)
+
+    scrollPadding2 (em 4) (px 2)
+
+    scrollPadding3 (em 4) (px 2) (pct 5)
+
+    scrollPadding4 (em 4) (px 2) (pct 5) (px 3)
+
+-}
```
BrianHicks: same comment here as on scrollMargin

src/Css.elm
brianvanburken: I've updated the documentation. I hope it clarifies a lot. I can add the same documentation to `margin` and `padding` if you want?

src/Css.elm
```diff
@@ -13849,6 +13859,7 @@ scrollMargin (Value value) =
 
 
 {-| Sets [`scroll-margin`](https://css-tricks.com/almanac/properties/s/scroll-margin/) property.
+_Note:_ See [`scrollMargin`](#scrollMargin) for documentation about the order of arguments.
```
BrianHicks: not to make you do a lot more work, but what would you think about documenting the positions only in the functions that use them? I think in some editors that will show up in autocomplete or docs-on-hover so it'd be way more useful in the functions themselves.

brianvanburken: I've updated the PR to add more documentation for each case. I'll add the same way of documentation to `padding` and `margin` functions in a new PR after this PR is approved

PR #508, opened 2019-10-08
Line height letter spacing

danielmcopeland:
- [x]  Each Value is an open record with a single field. The field's name is the value's name, and its type is Supported. For example foo : Value { provides | foo : Supported }
- [x]  Each function returning Style accepts a closed record of Supported fields.
- [x]  If a function returning Style takes a single Value, that Value should always support inherit, initial, and unset because all CSS properties support those three values! For example, borderFoo : Value { foo : Supported, bar : Supported, inherit : Supported, initial : Supported, unset : Supported } -> Style
- [x]  If a function returning Style takes more than one Value, however, then none of its arguments should support inherit, initial, or unset, because these can never be used in conjunction with other values! For example, border-radius: 5px 5px; is valid CSS, border-radius: inherit; is valid CSS, but border-radius: 5px inherit; is invalid CSS. To reflect this, borderRadius : Value { ... } -> Style must have inherit : Supported in its record, but borderRadius2 : Value { ... } -> Value { ... } -> Style must not have inherit : Supported in either argument's record. If a user wants to get border-radius: inherit, they must call borderRadius, not borderRadius2!
- [x]  When accepting a numeric Value (e.g. a length like px, an angle like deg, or a unitless number like int or num), always include zero : Supported as well as calc : Supported!
- [x]  Every exposed value has documentation which includes at least 1 code sample.
- [x]  Documentation links to to a CSS Tricks article if available, and MDN if not.
- [x]  Make a pull request against the phantom-types branch, not master!

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -38,7 +38,7 @@ module Css exposing
     , margin, margin2, margin3, margin4, marginTop, marginRight, marginBottom, marginLeft
     , boxSizing
     , alignItems, alignSelf
-    , fontSize, xxSmall, xSmall, small, medium, large, xLarge, xxLarge, smaller, larger
+    , fontSize, xxSmall, xSmall, small, medium, large, xLarge, xxLarge, smaller, larger, lineHeight, letterSpacing
     , fontFamily, fontFamilies, serif, sansSerif, monospace, cursive, fantasy, systemUi
     , fontStyle, italic, oblique
     , fontWeight, bold, lighter, bolder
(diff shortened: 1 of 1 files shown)


brianvanburken:
Thank you for contributing to the phantom type branch :heart:. One nitpick: `letterSpacing` has already been added in #505

BrianHicks:
hi Brian, that one's on me—this was a PR from a local Elm meetup at which I led a contribution workshop. I didn't include the things in your omnibus PR because I forgot to resolve conflicts before calculating the open properties!

If it's OK with you, I'm going to merge this one and we can improve it in your branch. If you could resolve the conflicts against the latest phantom-types I'm happy to review+merge it when it's clear to merge again. 😄

brianvanburken:
@BrianHicks sure thing! I've rebase the branch and it's free from conflicts and all green. So, you're good to go! Do you want me to revert my commit containing the addition of `letterSpacing`?

BrianHicks:
that'd be wonderful, thank you.

PR #515, opened 2019-12-18
Replace (int 1.5) with (num 1.5) to match type

Spaxe:
There are 4 instances in which `(int 1.5)` is used, but the Elm compiler will fail it. Replacing `int` with `num`  produces the correct unitless example.


Spaxe:
Oops. #492 already did it.

PR #517, opened 2020-03-16
Phantom Types: spec fixes

brianvanburken:
This PR fixes some of the difference between the official spec and our implementation

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -5593,6 +5593,7 @@ backgroundSize :
         , vmax : Supported
         , auto : Supported
         , cover : Supported
+        , contain : Supported
         , inherit : Supported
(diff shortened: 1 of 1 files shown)


brianvanburken:
@rtfeldman this PR fails when diffing the documentation telling me that `"I got the data back, but it was not what I was expecting. The response body contains 400171 bytes."`. Maybe it exceeds a limit?

leifhelm:
If is cause by `elm diff`. Try `elm diff 16.0.0 > /dev/null` to see if it throws any errors. It is used to check if the documentation is correct I guess.

brianvanburken:
This PR is combined in #538 so it can be closed.

PR #519, opened 2020-04-25
Expose CalculatedLength

jgrenat:
Hello!

A small PR to expose CalculatedLength in order to have type annotations if we need to.
For example, I'm exposing from my design system the remaining height of the page minus the header height:

```elm
remainingHeight =
    calc (pct 100) minus (px 100)
```


rtfeldman:
Makes sense - thanks @jgrenat!

PR #526, opened 2020-09-02
update travis.yml to elm 0.19.1

harrysarson:
Should remove the red crosses from CI!


rtfeldman:
Nice, thanks for fixing this!

PR #536, opened 2021-03-21
Phantom types: Fix bugs and add documentation

leifhelm:
Fixing bugs that I have detected by the use of a modified test-suite.
Added a bit of documentation along the way.


rtfeldman (approved):
Thanks @leifhelm! 🎉 

I know tests aren't passing right now, but these all look correct from inspection, so I'm happy to merge as-is.

Issue #534, opened 2021-01-22
`Css.important` on `Css.batch` output only applies to last Css.Style


Chadtech:
Hello!

Today I ran code like this..
```elm
    [ Css.borderTopLeftRadius radiusSize
    , Css.borderTopRightRadius radiusSize
    , Css.borderBottomLeftRadius radiusSize
    , Css.borderBottomRightRadius radiusSize
    ]
        |> Css.batch
        |> Css.important
```
.. and the `important` was only applied to the last style.

I managed to reproduce it in this ellie: https://ellie-app.com/c8R5mP4Kvxsa1 , where `specialBorderRadius` should be applying a border radius to the top right and bottom left corners because of the `important`, but it is only applied to the top right corner.

I can see the code for why this is. There appears to be a `mapLastProperty` function that `Css.important` is using. 

I would be happy to make the PR for this. Just let me know!

PR #537, opened 2021-03-24
Use new mapProperties function in important

Chadtech:
This solves the problem I wrote up in #534


rtfeldman (approved):
Thanks @Chadtech!

rtfeldman:
Released as 16.1.1

PR #538, opened 2021-04-02
Phantom types: Add a bunch of properties and tests

leifhelm:
This PR includes the following new properties:
`animation-name`
`animation-duration`
`animation-timing-function`
`animation-iteration-count`
`animation-direction`
`animation-play-state`
`animation-delay`
`animation-fill-mode`
`align-content`
`justify-items`
`justify-self`
`font-feature-settings`
`outline`
`outline-width`
`outline-color`
`outline-style`
`outline-offset`
`text-indent`
`width` @brianvanburken 
`min-width` @brianvanburken
`max-width` @brianvanburken
`height`
`min-height`
`max-height`
`!important`
`text-shadow` @brianvanburken
`order` @rlefevre
`flex` @rlefevre
`flex-direction` @rlefevre
`flex-grow` @rlefevre
`flex-shrink` @rlefevre
`flex-wrap` @rlefevre
`flex-flow` @rlefevre
`list-style-type`
`list-style-position`
`list-style-image`

And their  values.

I have enabled the property tests again to check if my code works. (And I caught a few mistakes already)

I have hacked something to together to validate the test suite with the [W3C CSS Validation tool](https://jigsaw.w3.org/css-validator/) and it passed except for:
`pointer-events` and `text-rendering` as they are not stable yet.
`animation-name` does accept `<string>` but the validator failed ([spec](https://www.w3.org/TR/css-animations-1/#typedef-keyframes-name)).
`stops()` with the values `jump-start`, `jump-end`, `jump-none`, and `jump-both` ([spec](https://www.w3.org/TR/css-easing-1/#step-easing-functions)).


brianvanburken (approved):
Amazing work ❤️! Looks good to merge (I reviewed all your commits individually since it was too huge going through all of it at once 😉). I don't know if @rtfeldman wants to review it too, else I'll merge this. Those aliases are a great addition too and will make it easier to add properties with similair values!

rtfeldman:
Awesome! Thanks so much @leifhelm, and thanks @brianvanburken for reviewing! 😻

PR #541, opened 2021-10-04
Make animation-iteration-count work with infinite

### LINKED ISSUE #510: "infinite" in animationIterationCount does not work
I've got a type error whenever I try to define a styled div with an animation on it.

A simple reproduction :
```
styled div []
  [ animationName animationNameHere
  , animationDuration (sec 1)
  , animationIterationCount infinite
  ]
```

The compilator's error :
```
The 1st argument to `animationIterationCount` is not what I expect:

42|       , animationIterationCount infinite
                                    ^^^^^^^^
This `infinite` value is a:

    Infinite

But `animationIterationCount` needs the 1st argument to be:

    { compatible
        | number : Compatible, numberOrInfinite : Compatible, value : String
    }
```

TSFoster:
Fixes #510

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -526,14 +526,13 @@ type alias Number compatible =
 
 {-| -}
 type alias NumberOrInfinite compatible =
-    Number { compatible | numberOrInfinite : Compatible }
+    { compatible | value : String, numberOrInfinite : Compatible }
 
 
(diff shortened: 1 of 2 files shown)


rtfeldman (approved):
Thanks @TSFoster!

PR #544, opened 2021-11-05
Optimize string concatination.

robinheghan:
I recently setup an elm-css benchmark based on code from our production code base, and after profiling said benchmark, I discovered that string concatination and hashing is where elm-css spends most of its time.

This PR improves the performance of elm-css by combining several map+join and filter+join operations into a single operation. It also replaces certain `String.join ""` calls with normal string concatination using the ++ operator.

From my testing, this improves performance by 15-20% depending on the browser being run.

The benchmark I ran can be found here: https://github.com/robinheghan/elm-css-playground

Check out the `reduce-string-joins` branch to test this particular optimization in isolation.

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -482,6 +482,7 @@ deprecated or discouraged.
 import Css.Animations exposing (Keyframes)
 import Css.Internal exposing (getOverloadedProperty, lengthConverter, lengthForOverloadedProperty)
 import Css.Preprocess as Preprocess exposing (Style, unwrapSnippet)
+import Css.String as String
 import Css.Structure as Structure exposing (..)
 import Hex
 import String
@@ -974,8 +975,7 @@ combineLengths operation firstLength secondLength =
             [ String.fromFloat numericValue
             , firstLength.unitLabel
             ]
-                |> List.filter (not << String.isEmpty)
-                |> String.join ""
+                |> String.filterJoin (not << String.isEmpty) ""
     in
     { firstLength | value = value, numericValue = numericValue }
 
@@ -6299,7 +6299,7 @@ borderColor2 : ColorValue compatibleA -> ColorValue compatibleB -> Style
 borderColor2 c1 c2 =
     let
         value =
-            String.join " " [ c1.value, c2.value ]
+            c1.value ++ " " ++ c2.value
     in
     property "border-color" value
 
@@ -6692,8 +6692,7 @@ fontFeatureSettings { value } =
 fontFeatureSettingsList : List (FeatureTagValue a) -> Style
 fontFeatureSettingsList featureTagValues =
     featureTagValues
-        |> List.map .value
-        |> String.join ", "
+        |> String.mapJoin .value ", "
         |> property "font-feature-settings"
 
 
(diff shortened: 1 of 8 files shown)


src/Css.elm
```diff
@@ -974,8 +975,7 @@ combineLengths operation firstLength secondLength =
             [ String.fromFloat numericValue
             , firstLength.unitLabel
             ]
-                |> List.filter (not << String.isEmpty)
-                |> String.join ""
+                |> String.filterJoin (not << String.isEmpty) ""
```
gampleman, line 978: I'm squinting at this, but why is the filter even necessary here? Wouldn't joining empty strings just end up being exactly the same? Is that a perf optimization in itself?

robinheghan: That's a very good point!

rtfeldman: Yeah it looks like this whole thing can be `String.fromFloat numericValue ++ firstLength.unitLabel`.

I suspect the reason it wasn't in the first place is that it was copy/pasted and modified from something that was joining with newlines (instead of `""`), where doing the filter actually meant not printing unnecessary blank lines.

src/Css/Structure/Output.elm
```diff
@@ -1,18 +1,18 @@
...
-    , String.join "\n" (List.map importToString imports)
-    , String.join "\n" (List.map namespaceToString namespaces)
-    , String.join "\n\n" (List.map prettyPrintDeclaration declarations)
+    , String.mapJoin importToString "\n" imports
+    , String.mapJoin namespaceToString "\n" namespaces
+    , String.mapJoin prettyPrintDeclaration "\n\n" declarations
     ]
-        |> List.filter (not << String.isEmpty)
-        |> String.join "\n\n"
+        |> String.filterJoin (not << String.isEmpty) "\n\n"
```
rtfeldman, line 15: As I recall, the origins of this being a pretty print go back to the original `elm-css` being a build-time tool that generated actual `.css` files, so it was nice to see the newlines and performance was less of a concern.

Today I think we should prioritize speed over readability of generated CSS, so I suggest taking this a step further!

```suggestion
        |> String.concat
```

src/Css.elm
rtfeldman: I think between this change and https://github.com/rtfeldman/elm-css/pull/544/files#r744673432 we don't need `filterJoin` and can take it out of the PR! 😄

rtfeldman (changes requested):
This is looking good - thanks so much for this @robinheghan!

A couple of changes I'd like to see:

* I think we can [do without `filterJoin`](https://github.com/rtfeldman/elm-css/pull/544/files#r744673792)
* All the `mapJoin` calls that are joining newlines can become `String.concat` I think. Should be even faster, and I don't think it should break any output!
* I'd rather we do `import Css.String exposing (mapJoin)` than `import Css.String as String` - I know people have different preferences on this, but I prefer not to override core modules 😄

robinheghan:
I believe this branch is ready to be merged.

I haven't changed `mapJoin` calls to `String.concat`, because I'm very certain it would be slower. `String.concat` does a `String.join ""`, which again converts the List to a JsArray. It would also require that we'd perform the `map` as a seperate action, which performs a `List.foldr` under the hood.

`mapJoin` iterates the list directly, which only requires iterating the list once.

rtfeldman (approved):
Cool, makes sense! 👍

PR #546, opened 2021-11-12
Use fnv hash

robinheghan:
This replaces the current Murmur3 based hash function with a FNV-based hash function.

Rationale can be found here: https://github.com/robinheghan/fnv1a/blob/main/docs/why_fnv_is_great_for_elm.md

Modest speedup in Chrome (~5%) while Safari and Firefox see huge increase in performance. With all my changes (past PRs included) Safari jumps from ~40% improvement to 55-60% improvement. FF jumps from ~35% to ~50%.


rtfeldman (approved):
Awesome, thanks! 🚀

PR #547, opened 2021-11-15
Add paddingBlockStart and paddingBlockEnd properties

dzuk-mutant:
I noticed these were missing from the package. They accept the same length values as padding-left/right/bottom/etc ^^

https://developer.mozilla.org/en-US/docs/Web/CSS/padding-block-start
https://developer.mozilla.org/en-US/docs/Web/CSS/padding-block-end


dzuk-mutant:
(In case this isn't fixed by the time someone who has control of repo sees it, GitHub seems to be having a bug with dummy email addresses right now, so the commits in this PR have been tagged as being made by a user who is not me. This is not intentional 😓)

dzuk-mutant:
Realised I should have put these on a different branch (and there were some more properties I could have added).

PR #548, opened 2021-11-16
Placeholder pseudo-element

dzuk-mutant:
https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder

While the MDN article mentions there are accessibility concerns in using this, as long as you understand the constraints and work with/counteract default browser styling it can be a useful part of designing inputs IMO.

If this would be an obstacle to including it in elm-css, perhaps the docs can mention that it's something that designers need to be aware of these sorts of issues if they want to use it.

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -7593,6 +7593,13 @@ firstLine : List Style -> Style
 firstLine =
     pseudoElement "first-line"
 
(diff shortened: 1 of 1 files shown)


dzuk-mutant:
(In case this isn't fixed by the time someone who has control of repo sees it, GitHub seems to be having a bug with dummy email addresses right now, so the commits in this PR have been tagged as being made by a user who is not me. This is not intentional 😓)

PR #549, opened 2021-11-16
Add padding-block properties

dzuk-mutant:
Adds properties from the `padding-block` group for creating paddings that are relative to reading direction. (https://developer.mozilla.org/en-US/docs/Web/CSS/padding-block)

```
paddingBlock (px10)

paddingBlock2 (px 10) (px 10)

paddingBlockStart (px 10)

paddingBlockEnd (px 10)
```

(In case this isn't fixed by the time someone who has control of this repo sees it, GitHub seems to be having a bug with dummy email addresses right now, so one of these commits in this PR have been tagged as being made by a user who is not me but happened to use the same dummy email in the past. This is not intentional, I do not have control of that GitHub account 😓)


Confidenceman02:
@rtfeldman Any reason this was not merged? Would be a nice one to have IMO.

PR #550, opened 2021-11-16
Phantom types: Logical paddings and margins

dzuk-mutant:
These were either not on the checklists or self-assigned by someone else as far as I could see so I figured it would be okay to make a PR request for them ^^ (Reading some more I think someone removed the padding ones because they used to be tagged as experimental, they are no longer considered experimental.)

Adds implementations for:
- `padding-block` (including 1 and 2-argument variants)
- `padding-block-start`
- `padding-block-end`
- `padding-inline` (including 1 and 2-argument variants)
- `padding-inline-start`
- `padding-inline-end`
- `margin-block` (including 1 and 2-argument variants)
- `margin-block-start`
- `margin-block-end`
- `margin-inline` (including 1 and 2-argument variants)
- `margin-inline-start`
- `margin-inline-end`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,16 @@
 # elm-css
 
+## Quick note
+
+This is a fork of rtfeldman's elm-css module. The differences between this one and the original is:
+
+- This uses the `phantom-types` branch. In short, it means you get much better compiler errors.
+- This has been updated with much newer properties. *However, no experimental properties and values have been included.*
+
+
+---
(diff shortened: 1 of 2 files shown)


dzuk-mutant:
Closing the pull request given the current state of elm-css and the fact that with this branch, I am now planning to do my own fork and have basically revamped the entire file at this point ^^

PR #556, opened 2021-12-03
Replace fnv with murmur3

robinheghan:
After talking to Richard, we agreed to hold back the hash-function change. FNV is a slightly worse hash than Murmur3, and elm-css doesn't have a way to handle hash collisions. It's safer to stick with Murmur3, for now.


rtfeldman (approved):
Lovely, thanks @robinheghan! 🎉

PR #559, opened 2022-01-07
Set display: none and add elm-css-style-wrapper class

### LINKED ISSUE #557: `<span> > <style>` wrapping in elm-css 17 has unexpected side effect inside grid container
In elm-css 17.0.1, `<style>` is wrapped in `<span>` to guard against some browser extension mess-ups.
The intention is good but it has unexpected side effect inside grid container.

SSCCE: https://ellie-app.com/g3z4zCTShLfa1

This example uses following style on container `div`:

```css
._5c4fac34 {
    position:fixed;
    top:0px;
    right:0px;
    bottom:0px;
    left:0px;
    width:100%;
    height:100%;
    display:grid;
    align-items:center;
    justify-items:center;
}
```

This is a convenient "just stay center!" container style, which can horizontally/vertically center its direct child via `align/justify-items: center`.

We were using this for creating "backdrop" area of modal dialogs and working pretty well.

However with elm-css 17.0.1, such container now hosts an additional child which is `<span>` containing style node, and that `<span>` now takes up a portion of the grid container since it is (at a glance from grid container) an ordinary element!

Screenshot from SSCCE:

image

A quick workaround is to put `display: none;` to wrapping `<span>`s. At least it works for this particular example, tested from dev console:

image

bcardiff:
Fixes #557

Using `display: none;` seems like a good default and still fixes #542.

I think it might be worth adding a class like `elm-css-style-wrapper` in case the display: none is not good or if someone needs to override the style as @ymtszw needed.

If that's the case using the following will allow to override even the inline `display: none;` [^1]

```css
.elm-css-style-wrapper[style] { 
   /* your style */
}
```

I based this change on 17.0.1 in case a 17.0.2 is wanted.

[^1]: [Override Inline Styles with CSS](https://css-tricks.com/override-inline-styles-with-css/)

### DIFF
--- a/src/Css/Global.elm
+++ b/src/Css/Global.elm
@@ -106,7 +106,10 @@ global snippets =
         |> VirtualDom.node "style" []
         -- wrap the style node in a div to prevent `Dark Reader` from blowin up the dom.
         |> List.singleton
-        |> VirtualDom.node "span" []
+        |> VirtualDom.node "span"
+            [ VirtualDom.attribute "style" "display: none;"
+            , VirtualDom.attribute "class" "elm-css-style-wrapper"
+            ]
         |> VirtualDom.Styled.unstyledNode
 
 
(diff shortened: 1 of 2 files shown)


rtfeldman:
Awesome, thanks @bcardiff! 🎉

PR #561, opened 2022-01-09
fix: Runtime exception when using css in SVG elements

### LINKED ISSUE #543: Runtime type error when using CSS in an SVG element's attributes
The following code compiles but throws `Uncaught TypeError: Cannot set property className of #<SVGElement> which has only a getter` in Chrome (and a similar error in Firefox):

```
module Main exposing (main)

import Html.Styled exposing (toUnstyled)
import Html.Styled.Attributes exposing (css)
import Svg.Styled exposing (svg)

main =
    svg [ css [] ] []
        |> toUnstyled
```

Ellie: https://ellie-app.com/fCySBLMNKcWa1

There is no error when importing the `css` function from `Svg.Styled.Attributes`.

Confidenceman02:
fixes #543 
# Context
When `unStyling` nodes elm-css internally was using `VirtualDom.property "className" ...` which was causing runtime exceptions for svg elements. The property "className" is deprecated for svg elements and was the reason for the exception.

This meant that anytime you would set css inside of any svg element, an exception would be thrown and cause the page to not render correctly. 

virtualdomerror

# Fix
Using `VirtualDom.attribute "className" ...` properly sets the value and stops the exception from happening.

# Extra
- Updated elm examples
- Replaced dead package
- Updated elm-test

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -25,7 +25,7 @@
             "Css.Global"
         ]
     },
-    "elm-version": "0.19.1 <= v < 0.20.0",
+    "elm-version": "0.19.0 <= v < 0.20.0",
     "dependencies": {
         "elm/core": "1.0.0 <= v < 2.0.0",
         "elm/json": "1.0.0 <= v < 2.0.0",
--- a/elm.json
+++ b/elm.json
@@ -25,7 +25,7 @@
             "Css.Global"
         ]
     },
-    "elm-version": "0.19.0 <= v < 0.20.0",
+    "elm-version": "0.19.1 <= v < 0.20.0",
     "dependencies": {
         "elm/core": "1.0.0 <= v < 2.0.0",
         "elm/json": "1.0.0 <= v < 2.0.0",
--- a/examples/html-styled/elm.json
+++ b/examples/html-styled/elm.json
@@ -4,7 +4,7 @@
         ".",
         "../../src"
     ],
-    "elm-version": "0.19.1",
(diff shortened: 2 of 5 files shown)


elm.json
```diff
@@ -25,7 +25,7 @@
             "Css.Global"
         ]
     },
-    "elm-version": "0.19.0 <= v < 0.20.0",
+    "elm-version": "0.19.1 <= v < 0.20.0",
```
rtfeldman, line 28: Is 0.19.1 required for this to work? (If not, I'd rather keep it compatible with 0.19.0!)

examples/html-styled/elm.json
```diff
@@ -4,29 +4,29 @@
         ".",
         "../../src"
     ],
-    "elm-version": "0.19.0",
+    "elm-version": "0.19.1",
```
rtfeldman, line 7: I assume this can be 0.19.0 too, yeah? 😄

elm.json
Confidenceman02: Not required. Back to 0.19.0

examples/html-styled/elm.json
Confidenceman02: :smile: Indeed! Reverted!

rtfeldman (approved):
Awesome, thanks for the quick fix! 😃

rtfeldman:
Published as 17.0.3!

Confidenceman02:
:raised_hands:

PR #562, opened 2022-01-09
fix: Set attribute value to class

### LINKED ISSUE #563: Css classes not rendering correctly
# Context
As as result of fixing #543 I introduced another probably more severe bug. Using className as the value to VirtualDom.attrbute incorrectly renders the classname attribute.

For non svg dom nodes it renders as `classname` and for svg nodes is renders as `className`. Both are incorrect syntax and means no styles are actually applied.

classnames

# Fix 
Using `class` fixes this issue.

Confidenceman02:
fixes #563 

# Context
Using attribute does set the css class correctly however the attribute
name needs to be class instead of className.

Using className will set classes as `classname` in dom nodes and
`className` in svg nodes which is incorrect.

Here we can see the incorrect attribute name. On the left we should be seeing a handsome `elm-select`. No styles are applied.
classnames

And with correct class.
classnamefixed

### DIFF
--- a/src/VirtualDom/Styled.elm
+++ b/src/VirtualDom/Styled.elm
@@ -492,10 +492,10 @@ extractUnstyledAttribute styles (Attribute val isCssStyles cssTemplate) =
     if isCssStyles then
         case Dict.get cssTemplate styles of
             Just classname ->
-                VirtualDom.attribute "className" classname
+                VirtualDom.attribute "class" classname
 
             Nothing ->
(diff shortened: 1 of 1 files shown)


Confidenceman02:
Many apologies for this oversight.

rtfeldman:
So just to confirm, `class` works with both SVG and HTML nodes?

rtfeldman:
Published as 17.0.4

Issue #503, opened 2019-07-11
Using `Html.Styled.Attributes`'s css breaks Svg.Styled at runtime


tesk9:
```
import Html.Styled.Attributes exposing (css)
import Svg.Styled as Svg exposing (Svg)

main =
    Svg.svg
        [ css [] ]
        []
        |> Svg.toUnstyled
```
throws an error: `Uncaught TypeError: Cannot assign to read only property 'className' of object '#<SVGSVGElement>'`

See https://ellie-app.com/64vQQmzM94Ca1

1000000000:
I think this is probably related to the following issues in the elm compiler and elm/svg:
* [elm/svg#3](https://github.com/elm/svg/issues/3)
* [elm/compiler#2060](https://github.com/elm/compiler/issues/2060)
* [elm/compiler#2032](https://github.com/elm/compiler/issues/2032)

PR #565, opened 2022-01-12
fix: setting styles with attribute for svg elements

### LINKED ISSUE #564: 17.0.4 generated class name is not added when there's also a set class name
(Example available at https://ellie-app.com/gnGZhWfMXcMa1.)

When there's just `css`, the styles work as expected:

```
import Css exposing (..)
import Html.Styled exposing (..)
import Html.Styled.Attributes exposing (class, css)


main =
    div
        [ css
            [ height (Css.px 100)
            , width (Css.px 100)
            , backgroundColor (rgb 100 0 0)
            , color  (rgb 100 0 0)
            ]
        ]
        [ text "This text should be hidden!" ]
        |> toUnstyled
```

And the generated class name is attached to the element:
Screen Shot 2022-01-11 at 9 27 05 AM

However, when a `class` attribute is also added,
```

import Css exposing (..)
import Html.Styled exposing (..)
import Html.Styled.Attributes exposing (class, css)


main =
    div
        [ css
            [ height (Css.px 100)
            , width (Css.px 100)
            , backgroundColor (rgb 100 0 0)
            , color  (rgb 100 0 0)
            ]
        , class "some-custom-class"
        ]
        [ text "This text should be hidden!" ]
        |> toUnstyled
```
The generated class name is not attached to the element:
Screen Shot 2022-01-11 at 9 26 41 AM

Which results in the styles not being applied.

Confidenceman02:
fixes #564 

# Context
17.0.4 fixed a bug that was causing DOM exceptions by changing the way we were updating styles to use `VirtualDom.attribute "class" ...` instead of `VirtualDom.property "className" ...`.

This was due to svg elements not liking the `property "className"` method which is a known phenomenon in the html spec.

Some random stack overflow post that is probably true: https://stackoverflow.com/questions/43750395/how-to-add-a-class-in-a-svg-element-circle-using-javascript

However, whilst the pesky exception was fixed there were some unintended side effects you can read about in this issue #564.

# Work completed
- Revert the `extractUnstyledAttribute` function to behave as it used to for html elements.
- Create an `extractUnstyledAttributeNS` function to set styles for svg nodes using the `attribute` method. 

The above changes focus on only fixing what was breaking, rather than changing the implementation for all elements.

Styled examples showing svg and html elements with their respective class properties. 

NS

### DIFF
--- a/examples/html-styled/Main.elm
+++ b/examples/html-styled/Main.elm
@@ -4,7 +4,7 @@ import Browser
 import Css exposing (..)
 import Html
 import Html.Styled exposing (..)
-import Html.Styled.Attributes exposing (css, href, src)
+import Html.Styled.Attributes exposing (class, css, href, src)
 import Html.Styled.Events exposing (onClick)
 import Svg.Styled as StyledSvg
 import Svg.Styled.Attributes as SvgAttribs
@@ -82,9 +82,9 @@ legacyBorderRadius amount =
 view : Model -> Html Msg
 view model =
     nav []
-        [ img [ src "assets/backdrop.jpg", css [ width (pct 100) ] ] []
+        [ img [ src "assets/backdrop.jpg", css [ width (pct 100) ], class "some-whatever-class" ] []
         , btn [ onClick DoSomething ] [ text "Click me!" ]
-        , StyledSvg.svg [ SvgAttribs.css [], SvgAttribs.width "100", SvgAttribs.height "100" ] [ StyledSvg.circle [ SvgAttribs.cx "50", SvgAttribs.cy "50", SvgAttribs.r "40" ] [] ]
+        , StyledSvg.svg [ SvgAttribs.class "some-whatever-NS-class", SvgAttribs.css [ opacity (num 0.5) ], SvgAttribs.width "100", SvgAttribs.height "100" ] [ StyledSvg.circle [ SvgAttribs.cx "50", SvgAttribs.cy "50", SvgAttribs.r "40" ] [] ]
         ]
 
 
--- a/examples/html-styled/Main.elm
+++ b/examples/html-styled/Main.elm
@@ -84,7 +84,13 @@ view model =
     nav []
         [ img [ src "assets/backdrop.jpg", css [ width (pct 100) ], class "some-whatever-class" ] []
         , btn [ onClick DoSomething ] [ text "Click me!" ]
-        , StyledSvg.svg [ SvgAttribs.class "some-whatever-NS-class", SvgAttribs.css [ opacity (num 0.5) ], SvgAttribs.width "100", SvgAttribs.height "100" ] [ StyledSvg.circle [ SvgAttribs.cx "50", SvgAttribs.cy "50", SvgAttribs.r "40" ] [] ]
+        , StyledSvg.svg
+            [ SvgAttribs.class "some-whatever-NS-class"
+            , SvgAttribs.css [ color (rgb 256 0 0)
+            , SvgAttribs.width "100"
+            , SvgAttribs.height "100"
+            ]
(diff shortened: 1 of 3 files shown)


examples/html-styled/Main.elm
```diff
@@ -82,9 +82,9 @@ legacyBorderRadius amount =
 view : Model -> Html Msg
 view model =
     nav []
-        [ img [ src "assets/backdrop.jpg", css [ width (pct 100) ] ] []
+        [ img [ src "assets/backdrop.jpg", css [ width (pct 100) ], class "some-whatever-class" ] []
         , btn [ onClick DoSomething ] [ text "Click me!" ]
-        , StyledSvg.svg [ SvgAttribs.css [], SvgAttribs.width "100", SvgAttribs.height "100" ] [ StyledSvg.circle [ SvgAttribs.cx "50", SvgAttribs.cy "50", SvgAttribs.r "40" ] [] ]
+        , StyledSvg.svg [ SvgAttribs.class "some-whatever-NS-class", SvgAttribs.css [ opacity (num 0.5) ], SvgAttribs.width "100", SvgAttribs.height "100" ] [ StyledSvg.circle [ SvgAttribs.cx "50", SvgAttribs.cy "50", SvgAttribs.r "40" ] [] ]
```
tesk9, line 87: The circle is a bit low-contrast against the background.

How would you feel about using bright red instead?

```suggestion
         , StyledSvg.svg
            [ class "some-whatever-NS-class"
            , SvgAttribs.css [ color (rgb 256 0 0) ]
            , SvgAttribs.width "100"
            , SvgAttribs.height "100"
            ]
            [ StyledSvg.circle
                [ SvgAttribs.cx "50"
                , SvgAttribs.cy "50"
                , SvgAttribs.r "40"
                , SvgAttribs.fill "currentcolor"
                ]
                []
            ]
```

tesk9 (commented):
I made a PR adding a couple of tests against this PR -- hope that's alright with you!

This approach looks reasonable to me as long as:
- we're sure that namespaced nodes will only ever be SVG nodes
- we are fine with a runtime exception occurring if a user every accidentally uses `Html.Styled.Attributes.class` with an `Svg.Styled` node (https://github.com/rtfeldman/elm-css/issues/503 doesn't want to get closed, I guess!)
- we are fine with elm-test's html helpers not being able to test SVG classes (for this one, I think we either need to fix it in elm-test, or we _have_ to be fine with it)

examples/html-styled/Main.elm
Confidenceman02: This is fantastic, many thanks for opening the PR I shall review and merge post haste. 

On your points -

- The changes I made definitely only ever deal with NodeNS and KeyedNodeNS types so I think we are good there.
We are limited in testing this in the package with elm-test but I don't see why we couldn't test with something like  [playwright](https://playwright.dev/). I did this for elm-select and it works great. Perhaps a little overkill but we could raise an issue to explore how to test the svg class behaviour.

- The current runtime exception regarding the svg styles already exists so whilst we haven't fixed the issue, we haven't made it worse.

- I think it makes sense to patch elm-test to handle svg classes.

tesk9: Yay, glad you like it!

- I haven't ever used playwright -- looks great! Maybe [Percy](https://percy.io/) could be neat to have in place too? 

- True! 🙈 

- No, as far as I can tell, I don't think elm-test with SVG classes was ever working 🤔

Confidenceman02: Looks like an issue in elm-test exists https://github.com/elm-explorations/test/issues/134. 

Had a super quick look at the project and I think a fix is relatively straightforwards. (I will regret saying that)

rtfeldman (approved):
This looks good to me! Thanks for the review and the extra improvements @tesk9!

@Confidenceman02 are you ready for merge and release, or want to make any final changes?

Confidenceman02:
Happy to merge and release @rtfeldman.

PR #566, opened 2022-01-13
Allow multiple css attributes on root node

### LINKED ISSUE #558: Different behavior for multiple`css` attributes between root of tree and children
I noticed some strange behavior that I'm pretty sure should be considered a bug. 

If I specify multiple `css` attributes on the root node of a DOM tree that will get passed to `toUnstyled` then only the last call to `css` gets applied.  See the following example: 

```
main =
    Browser.sandbox { init = (), update = \_ -> identity, view = view >> toUnstyled }

view : () -> Html Msg
view _ =
    div
        [ css [ backgroundColor (hex "0000FF") ]
        , css [ color (hex "FF0000") ]
        ]
        [ text "asdf" ]
```

This produces text that is red, but does not have a blue background. 

image

Also it's interesting here that two class names are produced, but only one has any css styles associated with it: 

image

If, however, we apply the `css` attributes to a child node instead of the root we see both styles applied: 

```
view : () -> Html Msg
view _ =
    div []
        [ div
            [ css [ backgroundColor (hex "0000FF") ]
            , css [ color (hex "FF0000") ]
            ]
            [ text "asdf" ]
        ]
```

image

Which resolution is correct? 

There are a couple comments in VirtualDom.Styled.elm that suggest that the overriding behavior is intentional: 

```
-- NOTE: Styles should always result in a classname, because they
-- can be detected later.
-- This way img [ css [ foo bar ], css [] ] wipes out the styles
-- as expected. (The latter will generate a classname of "_unstyled")
```

```
-- NOTE: Do String.isEmpty classname and not List.isEmpty styles
-- so that img [ css [ foo bar ], css [] ] wipes out the styles
-- as expected. (The latter will generate a classname of "_unstyled")
```

However, we seem to rely on being able to apply multiple `css` attributes quite heavily at NoRedInk. This seems to be the more useful behavior.

micahhahn:
Fix #558

After discussion on #558 with @rtfeldman we decided that the cat is out of the bag on restricting users from being able to apply multiple `css` attributes to a single node. This has already been allowed for child nodes of the node passed to `toUnstyled` so this PR simply allows this on the root node as well (and we get to pull some code 😁). 

This will cause less confusion for users of this library.  Previously converting a node to be lazy could cause `css` styles to be dropped.

### DIFF
--- a/src/VirtualDom/Styled.elm
+++ b/src/VirtualDom/Styled.elm
@@ -304,7 +304,7 @@ unstyleNS :
 unstyleNS ns elemType properties children =
     let
         initialStyles =
-            stylesFromProperties properties
+            List.foldl accumulateStyles Dict.empty properties
 
         ( childNodes, styles ) =
             List.foldl accumulateStyledHtml
@@ -331,7 +331,7 @@ unstyle :
 unstyle elemType properties children =
     let
         initialStyles =
-            stylesFromProperties properties
+            List.foldl accumulateStyles Dict.empty properties
 
         ( childNodes, styles ) =
             List.foldl accumulateStyledHtml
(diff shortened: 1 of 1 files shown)


rtfeldman:
Thanks @micahhahn!

rtfeldman:
Published as 17.1.1

PR #570, opened 2022-03-20
Add toNonceUnstyled functions for CSP nonce support

### LINKED ISSUE #569: CSP nonce support (to be able to protect against XSS attacks)
My company uses elm-css extensively. We need to be able to support CSP without unsafe-inline, in order to be security compliant. Elm-css generates inlined css-styles, which are blocked by the CSP-policy we need to implement.

There is a way to get around this, using [nonces](https://content-security-policy.com/nonce/). A nonce is a per-request cryptographically secure random value that is set as an attribute in the <style> tag. 

One way to go about this would be to add a function similar to [toUnstyled](https://package.elm-lang.org/packages/rtfeldman/elm-css/latest/Html-Styled#toUnstyled) which also accepts a nonce. It's unclear to me what the best way to pass the nonce would be. This would turn the autogenerated elm-css style tag from <style> {elm-css-stuff} </style>, to <style nonce={passed nonce}>{elm-css-stuff}</style>.

jfmengels:
Fixes #569

Happy to make any changes, and to take suggestions for better documentation of the function :slightly_smiling_face:

### DIFF
--- a/src/Html/Styled.elm
+++ b/src/Html/Styled.elm
@@ -1,5 +1,5 @@
 module Html.Styled exposing
-    ( styled, fromUnstyled, toUnstyled
+    ( styled, fromUnstyled, toUnstyled, toNonceUnstyled
     , Html, Attribute, text, node, map
     , h1, h2, h3, h4, h5, h6
     , div, p, hr, pre, blockquote
@@ -20,9 +20,9 @@ module Html.Styled exposing
     )
 
 {-| Drop-in replacement for the `Html` module from the `elm-lang/html` package.
-The only functions added are `styled`, `toUnstyled` and `fromUnstyled`:
+The only functions added are `styled`, `toUnstyled`, `toNonceUnstyled` and `fromUnstyled`:
 
-@docs styled, fromUnstyled, toUnstyled
+@docs styled, fromUnstyled, toUnstyled, toNonceUnstyled
 
 This file is organized roughly in order of popularity. The tags which you'd
 expect to use frequently will be closer to the top.
@@ -159,7 +159,7 @@ type alias Attribute msg =
             [ padding (px 30)
             , fontWeight bold
             ]
-    
+
     view : Model -> Html msg
     view model =
         [ text "These two buttons are identical:"
@@ -172,6 +172,7 @@ identical to the normal `button` function, except that it has pre-applied
 the attribute of `css [ padding (px 30), fontWeight bold ]`.
 You can pass more attributes to `bigButton` as usual (including other `css`
 attributes). They will be applied after the pre-applied styles.
+
 -}
 styled :
     (List (Attribute a) -> List (Html b) -> Html msg)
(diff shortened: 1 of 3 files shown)


jonathanmoregard:
Anything we can do to help get this in @rtfeldman ? :)

jfmengels:
@JohanWinther That depends on what `Options` is. If it's a record, you'd call it like
```elm
Html.toUnstyledWith
  { nonce = Just "nonce id"
  ,  -- ...
  }
  (view model)
```

I think the API looks okay in that case, but it is not future-proof, because any addition/change you make to that record is a breaking change.

To make it future-proof, you can use the builder pattern:

```elm
Html.toUnstyledWith
  (Html.defaultOptions
    |> Html.withNonce "nonce id"
  )
  (model view)
```

The ergonomics become quite different, and not necessarily idiomatic to the rest of the package. In addition, the `Html.Styled` module would have a few additional out-of-place functions, or you move the options to a separate module.

The Html List pattern, with `toUnstyledWith : List Option -> Html msg -> VirtualDom.Node msg`, probably makes a lot more sense here:

```elm
Html.toUnstyledWith
  [ Html.nonce "nonce id"
  ]
  (model view)
```

But then the question becomes: What do we need to future proof for? I haven't followed feature requests for this repo enough, but this seems to be the first option since last the few years that `elm-css` has been released. So my gut feeling is that this is premature configuration and that we should wait until such a use-case comes around (or that someone mentions an existing one in thread). At that point, we can duplicate the `toUnstyled` functions for all possible permutations, or add one with options. A breaking change is also possibleand not necessarily out of question.

Anyway. If there is no use-case for a list of options, I don't think it's worth adding. But happy to add it if maintainers wish for it.

rtfeldman (approved):
Lovely, thanks @jfmengels! 🎉

choonkeat:
I think I'm missing something while trying to use [`Html.Styled.toNonceUnstyled`](https://package.elm-lang.org/packages/rtfeldman/elm-css/latest/Html-Styled#toNonceUnstyled). I'm still getting 

```
Content Security Policy: The page’s settings blocked the loading of a resource at inline (“style-src”)
```

and traced it down by modifying the elm output js

```diff
 	for (var key in attrs)
 	{
 		var value = attrs[key];
+		if (key === 'style') {
+			console.log(value !== 'undefined' ? 'setAttribute' : 'removeAttribute', key, value, domNode);
+			continue;
+		}
 		typeof value !== 'undefined'
 			? domNode.setAttribute(key, value)
 			: domNode.removeAttribute(key);
```

to see this single line logged

Screenshot 2022-06-18 at 4 45 34 PM

which is from https://github.com/rtfeldman/elm-css/pull/559 https://github.com/rtfeldman/elm-css/blob/c5ffa8c2ec47d7822ce91aeb268a2dd3d398c72f/src/Css/Global.elm#L108-L111 https://github.com/rtfeldman/elm-css/blob/c5ffa8c2ec47d7822ce91aeb268a2dd3d398c72f/src/VirtualDom/Styled.elm#L483-L486

How do I avoid that?

pawelpacana:
According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#violation_cases, inline style attributes (like in `<span style="display: none;" class="elm-css-style-wrapper">...</span>`) also violate `style-src` policy which does not allow `unsafe-inline`.

Current `Html.Styled.toNonceUnstyled` implementation is not enough as @choonkeat pointed out.

PR #571, opened 2022-03-20
Remove unused code

jfmengels:
I noticed that this argument was unused and could therefore be removed. I have no clue what the argument was meant for in the first place :shrug: 

There is a lot of dead code in this project (even excluding the deprecated folder). Would you care for a clean up PR?


rtfeldman (approved):
Fun, thanks @jfmengels!

PR #572, opened 2022-03-20
Make Css.Structure.Property a custom type

jfmengels:
> Ideally, this would be `type Property = Property String` - but in order to reduce allocations, we're doing it as a `type alias` until union types with one constructor get unboxed automatically.

Trying to realize your years-old wish. Feel free to ignore, I just couldn't help myself :sweat_smile:

### DIFF
--- a/src/Css.elm
+++ b/src/Css.elm
@@ -1162,12 +1162,12 @@ important =
 
 
 makeImportant : Property -> Property
-makeImportant str =
+makeImportant (Property str) =
     if String.endsWith " !important" (String.toLower str) then
-        str
+        Property str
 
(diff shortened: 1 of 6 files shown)


rtfeldman (approved):
Ha, nice! This is technically a breaking change, but I'm game for it. 😄

jfmengels:
Hmm, I didn't think it was a breaking change, because I (still) think this is only internal. I'm having a hard time telling in practice because `elm diff` doesn't work for me for this package.

But I guess if it's exposed, then it's probably best that people don't confuse it with `String`s anyway :man_shrugging:

PR #577, opened 2022-05-22
Fix documentation for rgb, rgba, hsl, hsla, hex

miniBill:
Remove extraneous parenthesis to fix documentation visualizaiton on packages.elm-lang.org


rtfeldman (approved):
Thanks @miniBill! 🎉

PR #578, opened 2022-06-09
Add support for `line-height: normal;`

caseyWebb:
https://developer.mozilla.org/en-US/docs/Web/CSS/line-height


rtfeldman (approved):
I never knew about this one - thanks @caseyWebb!

brian-carroll:
`lineHeight` used to be compatible with `int` and `zero` in previous versions. Probably still should, I think?

PR #585, opened 2022-09-08
Html styled keyed experiment

micahhahn:
This PR is an experiment on top of https://github.com/rtfeldman/elm-css/pull/584 where we join keyed + lazy together into one function. 

I feel pretty happy with the results but there were a couple snags where I had to make tradeoffs: 
- We need to eat another of the lazy args to pass the key, meaning we can only have `lazyNode` ... `lazyNode6`
- Elm does not support tuples above 3 elements, so the signatures of `lazyNode` to `lazyNode3` look different from `lazyNode4` to `lazyNode6`.  ["Once you get to four [tuple elements], it is best to make a record with named entries." 🤷‍♂️](https://github.com/elm/compiler/blob/master/hints/tuples.md)


micahhahn:
Closing as changes have been merged into #584

PR #586, opened 2022-12-05
Add gap

Chadtech:
Hello!

`gap` is a fairly new css property. I reach for it quite a lot. In flex containers with `gap`, all the children will have a gap of that size between them, getting the same effect as putting a `margin-top` on every element in a list except the first one (as developers sometimes do).


schoettl:
Fixes #587

mstruebing:
Can we please have this merged @rtfeldman ?

